Issue #11, opened 2017-06-17
Cleanup test files


ChrisRackauckas:
https://github.com/JuliaDiffEq/BoundaryValueDiffEq.jl/blob/master/test/runtests.jl#L7

These shooting method parts should probably be their own file, which is also included. 

Also,

https://github.com/JuliaDiffEq/BoundaryValueDiffEq.jl/blob/master/test/runtests.jl#L5

I prefer to wrap each included file into a `@testset`. This will make it run the entire testset, instead of erroring at the first failed test in the file. It also gives a very nice named printout to make it easier to follow what's going on. Example:

https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/test/runtests.jl

(though don't follow the `ode` folder thing: that's a relic of the past that needs to be dealt with... you can group related tests in folders, but you don't have enough tests for that to matter quite yet).

Issue #9, opened 2017-06-15
Abstract typing


ChrisRackauckas:
In the DiffEq-ecosystem, the solvers internally match the input type given by the user. In the BVP case, this is a little more complicated but for now let's keep the idea of `u0` for this purpose. If that's the case, you don't necessarily have a `Vector{Vector}`, but a `Vector{AbstractArray}` for some kind of `AbstractArray` determined by `u0`. 

It looks like you don't have much required to expand you typing, though these function signatures are overtyped:

https://github.com/JuliaDiffEq/BoundaryValueDiffEq.jl/blob/master/src/collocation.jl#L8

(and the `::Function` is unnecessary too)

https://github.com/JuliaDiffEq/BoundaryValueDiffEq.jl/blob/master/src/collocation.jl#L15

I don't see the reason to force type matching here. There can be very good reasons to use a separate type for independent and dependent variables. One case which comes to mind is using rationals for the independent variables. `dx=1//10` allows you to make sure you don't have the floating point issues with `dx=1/10` having `dx = 1/10 + 1/10 + ... + 1/10 = 0.9999994 != 1.0`, and using rationals here really doesn't have a computational cost. However, using rationals for the dependent variables can have a massive computational cost since those have to update, while the `x` are constant. That's a long explanation of: if you use the types the user gives you, it can be very useful to the user. 

https://github.com/JuliaDiffEq/BoundaryValueDiffEq.jl/blob/master/src/algorithms.jl#L19

That's another spot which assumes `Vector{Vector}`.

Issue #8, opened 2017-06-15
Clean Files


ChrisRackauckas:
https://github.com/JuliaDiffEq/BoundaryValueDiffEq.jl/blob/master/src/algorithms.jl#L17

These functions definitely don't belong in there.

Issue #6, opened 2017-06-15
Functions in Types


ChrisRackauckas:
Every function has its own type. But callable types are not `<:Function`. So you might want to let

https://github.com/JuliaDiffEq/BoundaryValueDiffEq.jl/blob/master/src/BoundaryValueDiffEq.jl#L38

those roam free without a restriction.

PR #13, opened 2017-06-18
[WIP] Files Clean up

YingboMa:
Fixing issues like #6 #8 #9 and #11 .

### DIFF
--- a/test/mirk_convergence_tests.jl
+++ b/test/mirk_convergence_tests.jl
@@ -1,3 +1,7 @@
+using BoundaryValueDiffEq
+using DiffEqBase, DiffEqDevTools
+using Base.Test
+
 # First order test
 function func_1!(x, y, du)
     du[1] = y[2]
--- a/test/mirk_convergence_tests.jl
+++ b/test/mirk_convergence_tests.jl
@@ -1,6 +1,3 @@
-using Base.Test
-using BoundaryValueDiffEq, DiffEqBase, DiffEqDevTools
-
 # First order test
 function func_1!(x, y, du)
     du[1] = y[2]
@@ -37,7 +34,7 @@ affineTol = 1e-9
 dts = 1./2.^(5:-1:1)
 order = 4
 
-println("Collocation method: MIRK")
+println("Collocation method (MIRK)")
 println("Affineness Test")
 prob = probArr[1]
 sol = solve(prob, MIRK(order), dt=0.2)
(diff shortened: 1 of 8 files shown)


test/mirk_convergence_tests.jl
```diff
@@ -1,6 +1,3 @@
-using Base.Test
-using BoundaryValueDiffEq, DiffEqBase, DiffEqDevTools
-
 # First order test
```
ChrisRackauckas: As a style thing, I prefer to have the `using` statements at the top of the test to make the script runnable. It's not necessary for `Pkg.test` since you'll `using` in `runtests` most of the stuff you need, but I like the idea that each test script is a runnable file that you can use to reproduce examples (i.e. every test file is also an example for how to use the package). Not mandatory, but something to keep in mind.

test/runtests.jl
```diff
@@ -1,35 +1,6 @@
...
-#  du[1] = 10.0(u[2]-u[1])
-#  du[2] = u[1]*(28.0-u[3]) - u[2]
-#  du[3] = u[1]*u[2] - (8/3)*u[3]
-# end
-
-# bc_lorenz(sol) = [sol[1][1]-1, sol[end][1]+10]
-# init_lorenz = [1.0,0.0,0.0]
-# lorenzprob = BVProblem(lorenz, (0.0,1.0), bc_lorenz, init_lorenz)
-# @test norm(bc_lorenz(solve(lorenzprob,Shooting(Tsit5())))) < 0.7
+@time @testset "Shooting Method Tests" begin include("shooting_tests.jl") end
```
ChrisRackauckas: I like to `tic() toc()` the whole thing to track timing changes. This gives you a general number to keep mind. It'll have small changes all of the time, but if you ever see a huge jump in how long your tests take, you'll be instantly noted that a performance regression happened.

test/shooting_tests.jl
```diff
@@ -0,0 +1,30 @@
...
+
+tspan = (0.,100.)
+u0 = [0.,1.]
+bvp = BVProblem(f, bc!, u0, tspan)
+resid_f = Array{Float64}(2)
+bc!(resid_f, solve(bvp, Shooting(Tsit5())))
+@test norm(resid_f) < 1e-7
+
+# function lorenz(t,u,du)
+#  du[1] = 10.0(u[2]-u[1])
```
ChrisRackauckas: Why is this test commented out?

YingboMa: Lorenz problem is too complicated for Shooting method, so I commented it out.

ChrisRackauckas: Should probably just remove it then. Replace it with the Lambert problem that's in the other file.

PR #14, opened 2017-06-20
[WIP] Optimize the MIRK method

YingboMa:
MIRK currently has huge amount of allocations and very slow. [Here](https://github.com/YingboMa/MIRK-profile) is the profile plot(s) of the following problem.

```
using BoundaryValueDiffEq
function fun!(x, u, du)
    du[1] = u[2]
    du[2] = -200*sin(u[1])
end

function boundary!(residual, u)
    residual[1] = u[1][1]
    residual[2] = u[end][1]+u[end÷2][1]+3+u[end÷2][2]
end

tspan = (0.,1.)
u0 = [0.,-2.]
dt = .02
bvp = BVProblem(fun!, boundary!, u0, tspan)
@profile sol = solve(bvp, MIRK(4), dt=dt)
```

### DIFF
--- a/src/BoundaryValueDiffEq.jl
+++ b/src/BoundaryValueDiffEq.jl
@@ -31,18 +31,19 @@ immutable MIRKTableau{T}
 end
 
 # ODE BVP problem system
-immutable BVPSystem{T}  # Order of the system
-    order::Int          # The order of MIRK method
-    M::Int              # Number of equations in the ODE system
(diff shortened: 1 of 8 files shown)


ChrisRackauckas:
Could you post the benchmark results between them? I am curious how much this changed the allocs, and think it should be documented.

YingboMa:
Ah... I mess up git.

PR #15, opened 2017-06-20
[WIP] Optimize the MIRK method

YingboMa:
MIRK currently has huge amount of allocations and very slow. [Here](https://github.com/YingboMa/MIRK-profile) is the profile plots and benchmark results of the following problem.

```
using BoundaryValueDiffEq
function fun!(x, u, du)
    du[1] = u[2]
    du[2] = -200*sin(u[1])
end

function boundary!(residual, u)
    residual[1] = u[1][1]
    residual[2] = u[end][1]+u[end÷2][1]+3+u[end÷2][2]
end

tspan = (0.,1.)
u0 = [0.,-2.]
dt = .02
bvp = BVProblem(fun!, boundary!, u0, tspan)
@profile sol = solve(bvp, MIRK(4), dt=dt)
```

### DIFF
--- a/src/BoundaryValueDiffEq.jl
+++ b/src/BoundaryValueDiffEq.jl
@@ -22,12 +22,12 @@ function BVProblem(f,bc,u0,tspan; iip = DiffEqBase.isinplace(f,3))
     BVProblem{typeof(u0),eltype(tspan),iip,typeof(f),typeof(bc)}(f,bc,u0,tspan)
 end
 
-immutable MIRKTableau{T}
+immutable MIRKTableau{T, U<:AbstractArray}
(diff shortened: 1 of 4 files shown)


YingboMa:
@ChrisRackauckas According to the [profile plot](https://github.com/YingboMa/MIRK-profile/blob/master/Third/profile_3.svg), the largest overhead are `flatten_vector` and `nest_vector` functions, so this might be a good point to start to discuss how do deal with the `NLsolve.jl` API.

ChrisRackauckas:
See what happens when you cache the matrix for `flatten_vector` and make that inplace.

PR #16, opened 2017-06-21
Use in-place vector of vectors operations

YingboMa:


YingboMa:
The latest profile and benchmark are [here](https://github.com/YingboMa/MIRK-profile/tree/master/PR16).

PR #32, opened 2018-01-19
Match with NLsolve new API

YingboMa:


REQUIRE
```diff
@@ -1,6 +1,6 @@
 julia 0.6
 DiffEqBase
 Reexport
-NLsolve
+NLsolve 0.12.1
```
ChrisRackauckas: The argument flip is not until v0.14.0? @pkofod

pkofod: Indeed

ChrisRackauckas:
Let's wait on @pkofod to tag, then let CI pass and pull this in.

pkofod:
I'm just waiting for metadata.

PR #33, opened 2018-01-28
Beasty blacksmith parameter shoot

BeastyBlacksmith:
haven't tested it yet, but I guess this is the functionality I am missing at the moment.
`DEFAULT_NLSOLVE_2` is not really necessary but it is more expressive this way imo, or we change the name of the second argument in `DEFAULT_NLSOVE`.

### DIFF
--- a/src/solve.jl
+++ b/src/solve.jl
@@ -24,6 +24,30 @@ function solve(prob::BVProblem, alg::Shooting; kwargs...)
     sol
 end
 
+function solve(prob::BVProblem, alg::ParameterShooting; kwargs...)
+    bc = prob.bc
+    u0 = deepcopy(prob.u0)
+    # Form a root finding function.
+    loss = function (resid, minimizer)
+        uEltype = eltype(minimizer)
+        tspan = (uEltype(prob.tspan[1]),uEltype(prob.tspan[2]))
+        tmp_prob = ODEProblem(prob.f,minimizer,tspan)
+        sol = solve(tmp_prob,alg.ode_alg;kwargs...)
+        bc(resid,sol,sol.prob.p,sol.t)
+        nothing
+    end
+    opt = alg.nlsolve(loss, prob.p)
+    sol_prob = ODEProblem(prob.f,opt[1],prob.tspan,prob.p)
+    sol = solve(sol_prob, alg.ode_alg;kwargs...)
+    if sol.retcode == opt[2]
+        solution_new_retcode(sol,:Success)
+    else
+        solution_new_retcode(sol,:Failure)
+    end
+    sol
+end
+
+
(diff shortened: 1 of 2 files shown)


src/solve.jl
```diff
@@ -24,6 +24,30 @@ function solve(prob::BVProblem, alg::Shooting; kwargs...)
...
 end
 
+function solve(prob::BVProblem, alg::ParameterShooting; kwargs...)
+    bc = prob.bc
+    u0 = deepcopy(prob.u0)
+    # Form a root finding function.
+    loss = function (resid, minimizer)
+        uEltype = eltype(minimizer)
+        tspan = (uEltype(prob.tspan[1]),uEltype(prob.tspan[2]))
+        tmp_prob = ODEProblem(prob.f,minimizer,tspan)
```
ChrisRackauckas: Need to pass the parameters

src/solve.jl
```diff
@@ -24,6 +24,30 @@ function solve(prob::BVProblem, alg::Shooting; kwargs...)
     sol
 end
 
+function solve(prob::BVProblem, alg::ParameterShooting; kwargs...)
+    bc = prob.bc
+    u0 = deepcopy(prob.u0)
+    # Form a root finding function.
+    loss = function (resid, minimizer)
```
ChrisRackauckas: The minimizer is the parameters, not the initial condition, so it needs to be placed correctly.

ChrisRackauckas:
I think that parameter shooting is better handled by the stuff in DiffEqParamEstim.jl (which does shooting) because those do minimizing instead of rootfinding, and with parameters/data there might not be a true minimum. It may be best to just add a BVP-specific loss function for `build_loss_objective`.

BeastyBlacksmith:
Well, my experience with numerical minimizing is varying. Somehow the outcome is not very predictable to me, thats why I tend to trust the rootfinding algorithms more.
So this approach might not be suited in general, plus I am still not quite sure how to update the initial conditions once the parameters have changed ( do I need a callback for this? ).

ChrisRackauckas:
>Well, my experience with numerical minimizing is varying. Somehow the outcome is not very predictable to me, thats why I tend to trust the rootfinding algorithms more.

Any rootfinding algorithm is the same as a minimization algorithm on the derivative. You can use trust-region Newton to do the minimization and will get a similar result, just without requiring the `f(x)=0` convergence which isn't guaranteed when parameters get involved. But I guess it can be kept as a separate thing, more related to data matching.

>I am still not quite sure how to update the initial conditions once the parameters have changed ( do I need a callback for this? ).

Make the minimizer array be both of them concatenated together, and split it into two views.

BeastyBlacksmith:
> Make the minimizer array be both of them concatenated together, and split it into two views.

I mean to have the initial conditions be a function of the parameters, that function has to be passed to the `solve` routine.

PR #53, opened 2019-12-01
update to `n+1` points in range

magerton:
We need `n+1` points if we have `n` steps of length `dt` from `t0` to `T` b/c we need to include `t0` in the range


magerton:
See output from Jupyter notebook below. `tspan = (0,5)` and with `dt=0.05`, but the time steps in the solution look like they are of length 5/99=0.050505... not 5/100=0.05

image

ChrisRackauckas:
Thanks!

PR #66, opened 2021-01-12
Fix different types for x and y

MatFi:
Proposal to fix error for complex valued u0 as mentioned in [Discourse](https://discourse.julialang.org/t/can-we-solve-complex-valued-two-point-boundary-value-problems/53232).
Removed the `tmp` from `mutable struct BVPSystem` because it  was never used.

### DIFF
--- a/src/collocation.jl
+++ b/src/collocation.jl
@@ -15,11 +15,11 @@ end
 
 # Dispatch aware of eltype(x) != eltype(prob.u0)
 function BVPSystem(prob::BVProblem, x, order)
-    Y = eltype(prob.u0)
+    T = eltype(prob.u0)
     M = length(prob.u0)
     N = size(x,1)
-    y = vector_alloc(Y, M, N)
-    BVPSystem(order, M, N, prob.f, prob.bc, prob.p, x, y, vector_alloc(Y, M, N), vector_alloc(Y, M, N), typeof(x)(undef,M))
+    y = vector_alloc(T, M, N)
+    BVPSystem(order, M, N, prob.f, prob.bc, prob.p, x, y, vector_alloc(T, M, N), vector_alloc(T, M, N), typeof(x)(undef,M))
 end
 
 # Auxiliary functions for evaluation
--- a/src/collocation.jl
+++ b/src/collocation.jl
@@ -10,16 +10,16 @@ end
 function BVPSystem(fun, bc, p, x, y, order)
     T, U = eltype(x), eltype(y)
     M, N = size(y)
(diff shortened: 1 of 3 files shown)


YingboMa (commented):
This is breaking. Maybe add a dispatch that has the signature `BVPSystem(::BVProblem, ...)`?

MatFi:
> This is breaking. Maybe add a dispatch that has the signature `BVPSystem(::BVProblem, ...)`?

Should be ok now

ChrisRackauckas:
@YingboMa why would it be breaking?

YingboMa:
In that commit, the function signature was changed.

src/collocation.jl
```diff
@@ -10,7 +10,16 @@ end
 function BVPSystem(fun, bc, p, x, y, order)
     T, U = eltype(x), eltype(y)
     M, N = size(y)
-    BVPSystem{T,U}(order, M, N, fun, bc, p, x, y, vector_alloc(T, M, N), vector_alloc(T, M, N), eltype(y)(M))
+    BVPSystem{T,U}(order, M, N, fun, bc, p, x, y, vector_alloc(U, M, N), vector_alloc(U, M, N), typeof(x)(M))
```
YingboMa, line 13: This is not right. It should be `vector_alloc(T, M, N)`. `U` is `Vector{T}`.

src/collocation.jl
```diff
@@ -10,7 +10,16 @@ end
...
+end
+
+# Dispatch aware of eltype(x) != eltype(prob.u0)
+function BVPSystem(prob::BVProblem, x, order)
+    U = eltype(prob.u0)
+    M = length(prob.u0)
+    N = size(x,1)
+    y = vector_alloc(U, M, N)
+    BVPSystem(order, M, N, prob.f, prob.bc, prob.p, x, y, vector_alloc(U, M, N), vector_alloc(U, M, N), typeof(x)(undef,M))
```
YingboMa, line 22: Please change the variable name from `U` to `T` to avoid confusion.

src/collocation.jl
MatFi: Oh now I see, this I messed up, I'll undo this.

src/collocation.jl
MatFi: I think that `T` is still confusing here, since it has always corresponded to `eltype(x)`, don't you think that one should choose something free, e.g. `Y` so that `U` is `Vector{Y}`, just to indicate that `Y ≠ T` is possible?

YingboMa: It's pretty standard to use `T` as the `eltype` of an array.

Issue #54, opened 2019-12-07
Adaptivity in MIRK solver


seadra:
(I'm not sure to what extent this overlaps with #12 which mentions defects)

The possibility of having this was mentioned in this [blog post](https://juliadiffeq.org/2019/06/06/StiffSDEs.html):

> Adaptivity in the MIRK BVP solvers

Lack of adaptivity strongly limits the usability in any real use. (For example, I currently rely on Mathematica for BVPs.)

Any hopes we will have adaptivity in a near future?

ChrisRackauckas:
The repo is setup for it, and the method itself can be adaptive, it's just no one ever got to it. If @yingboma finds time it might be a good time to finish it up. It just needs three steps: the definition of the interpolation, then the calculation of error estimates, then a resizing. It's not very difficult, but our main priority has been IVPs for awhile. As our work on stiff IVPs is starting to wrap up, this is a nice thing to keep in mind as an easy thing to finish up.

jibaneza:
This together with #27 would increase a lot the usability of this package. Or maybe it would easier to implement the ODEInterface BVP solvers within the DifferentialEquations package?

ChrisRackauckas:
Indeed. I am planning to look for covering it in the next GSoC round.

Issue #12, opened 2017-06-17
Defect control adaptivity


ChrisRackauckas:
http://epubs.siam.org/doi/pdf/10.1137/S1064827593251496
http://cs.stmarys.ca/~muir/ShampineMuirXu2006.pdf
http://epubs.siam.org/doi/pdf/10.1137/0912052

The next step should be adaptivity. A really nice way to handle the interpolations in a way that can be used internally is to simply add them to the solution object, and use that solution object on the inside. You can actually see this done in OrdinaryDiffEq.jl here:

https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/solve.jl#L250

So the steps for a defect control are:

1) Add a way to call an interpolation scheme on the `sol` type (which is algorithm-dependent). You might want to match OrdinaryDiffEq.jl for this, though it's quite complicated. We should discuss this.
2) Change the timing that the solution is built so it's built before the rest of the routine, and your solver routine just updates the `sol.u` and `sol.t` parts (this should actually require no change if the pointers are setup correctly!)
3) Then, the parts of the adaptive algorithm which require using the continuous extension for the defect, you can just call `sol(t)` (just like a user would).

The nice thing about this is that if you get the interpolation on the `sol` type right, then it'll work both internally and for users just the same. Also, this means you can pass the "current `sol`" to the `bc!`, making it exactly the same as the `Shooting` setup and allowing conditions on the interpolation.

PR #89, opened 2023-05-29
Add defect control adaptivity

ErikQQY:
Fix: #12 
Fix: #54 

Relating reference: https://netlib.org/ode/mirkdc.f

I just built the basic structure for the defect control adaptivity, it still needs a lot of work to get this program working.

The detailed explanation can be viewed on [my blog post](https://erikqqy.github.io/2023/05/29/preparation/)

ping @ChrisRackauckas  @YingboMa  @avik-pal


src/solve.jl
```diff
@@ -23,70 +23,446 @@ function DiffEqBase.__solve(prob::BVProblem, alg::Shooting; kwargs...)
...
-        Φ!(S, tableau, cache)
-        general_eval_bc_residual!(S)
-        flatten_vector!(resid, S.residual)
-        nothing
+```
+    redistribute(mesh_current, n, Nsub_star, s_hat)
+
+Generate a new mesh 
+
+```
```
avik-pal, line 213:
```suggestion
"""
    redistribute(mesh_current, n, Nsub_star, s_hat)

Generate a new mesh 

"""
```

avik-pal (changes requested):

src/solve.jl
avik-pal: Same for others below

src/solve.jl
```diff
@@ -17,71 +17,493 @@ function DiffEqBase.__solve(prob::BVProblem, alg::Shooting; kwargs...)
...
+            integral = 0
+        else
+            integral += next_piece
+            t = mesh_current[k + 1]
+            k += 1
+        end
+    end
+    mesh_new[Nsub_star + 1] = mesh_current[n + 1]
+    return mesh_new
+end
```
ChrisRackauckas, line 262: Can you describe what this one is doing?

src/solve.jl
```diff
@@ -17,71 +17,493 @@ function DiffEqBase.__solve(prob::BVProblem, alg::Shooting; kwargs...)
...
+    MxNsub = 3000
+
+    safety_factor = 1.3
+    rho = 1.0 # Set rho=1 means mesh distribution will take place everytime.
+    upper_new_mesh = 4.0
+    lower_new_mesh = 0.5
+    r1 = 0.0
+    r2 = 0.0
+    Nsub_star = 0
+    info = 0
```
ChrisRackauckas, line 180: are these pulled from the paper?

src/solve.jl
```diff
@@ -17,71 +17,493 @@ function DiffEqBase.__solve(prob::BVProblem, alg::Shooting; kwargs...)
...
+        if Nsub_star < lower_new_mesh * n
+            Nsub_star = lower_new_mesh * n
+        end
+        if Nsub_star > MxNsub
+            # Mesh redistribution fails
+            println("New mesh would be too large")
+            info = -1
+        else
+            println("Mesh redistributing")
+            mesh_new = redistribute(mesh_current, n, Nsub_star, s_hat)
```
ChrisRackauckas, line 221: I don't undersatnd why one would redistribute. Wouldn't this make it so you aren't clustering the points where the error is the most anymore?

src/solve.jl
```diff
@@ -17,71 +17,493 @@ function DiffEqBase.__solve(prob::BVProblem, alg::Shooting; kwargs...)
...
+
+        if opt[2] == ReturnCode.Success
+            info = 0
+        elseif opt[2] == ReturnCode.Failure
+            info = 1
+        end
+
+        # Change the original solution Vector{Vector{Float64}} to a matrix
+        # has size of (number of subintervals, length of equation)
+        global Y = transpose(reduce(hcat, S.y))
```
ChrisRackauckas, line 76: why is this global?

ErikQQY: The discrete solution ```Y``` is a local variable in the loop of iteration, so I set ```Y``` as global so we can finally build a solution based on ```Y```.

src/solve.jl
ErikQQY: From my understanding, after we redistribute the mesh and by using the new mesh to generate a new discrete solution, the new solution would have a smaller error.

src/solve.jl
ErikQQY: The BVP book does have the specification of r1, r2 and r3

image


But both the book and the paper don't elaborate on the setting of the other parameters, so I set these parameters according to the FORTRAN routine.

src/solve.jl
ErikQQY: The redistribute part generates a new mesh to make sure that on the next iteration, the new solution on the new mesh would have a smaller defect estimation.

Based on the judging quantities we computed before using defect and mesh function values, we compare the ratio $\frac{r1}{r3}$ with $\rho$ to decide whether we halve the mesh or make a redistribute.
```julia
    s_hat = zeros(Float64, n)
    for i in 1:n
        h = mesh_current[i + 1] - mesh_current[i]
        norm = abs(defect[i, idamax(defect[i, :])])
        s_hat[i] = (norm / tol)^(1.0 / (p + 1)) / h
        if s_hat[i] * h > r1
            r1 = s_hat[i] * h
        end
        r2 = r2 + s_hat[i] * h
    end
    r3 = r2 / n
    n_predict::Int64 = round(Int, (safety_factor * r2) + 1)
    if abs((n_predict - n) / n) < 0.1
        n_predict = round(Int, 1.1 * n)
    end
```
 If we need to redistribute(r1/r3>rho), we redistribute a new mesh, I keep some detailed notes on the redistribute process here: https://erikqqy.github.io/2023/06/04/gsoc-1st-week/#Redistribute-mesh

ChrisRackauckas:
What's left to complete this? Looks like there's some merge issues that are just from the formatter.

ErikQQY:
The main issue here is that the defect norm from the defect estimation part is always larger than the threshold, making the redistribution part hit an endless loop.

I think that the main problem can be located in the non-linear solving part of BundaryValueDiffEq.jl and MIRKDC, they are using different Newton iteration techniques, which leads to the difference in discrete stages ```K```, thus the defect we finally get maybe result in a different outcome.

src/cache.jl
```diff
@@ -17,22 +17,26 @@ function alg_cache{T,U}(alg::MIRK4, S::BVPSystem{T,U})
 end
 =#
 
-struct MIRK4GeneralCache{kType} <: GeneralMIRKCache
```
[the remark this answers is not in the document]
ErikQQY:
```julia
for i in 1:(N - 1)
    h = x[i + 1] - x[i]
    # Update K
    update_K!(S, cache, TU, i, h)
    for r in 1:s
        x_new = x[i] + c[r] * h
        y_new = (1 - v[r]) * y[i] + v[r] * y[i + 1]
        if r > 1
            y_new += h * sum(j -> X[r, j] * K[j], 1:(r - 1))
        end
        temp = zeros(Float64, M)
        fun!(temp, y_new, S.p, x_new)
        K[r] = temp[:]
    end
    # Update residual
    residual[i] = y[i + 1] - y[i] - h * sum(j -> b[j] * K[j], 1:s)
    cache.k_discrete[i, :] = K[:]
end
```

The MIRK cache stores the discrete stages from evaluating ```PHI``` in the nonlinear solving process. Since we only need discrete stages from the final nonlinear solving step, I think the previous steps can be omitted, but I didn't figure out an executable way, e.g. judging if this step is the last time we evaluate ```PHI```🤔.

src/adaptivity.jl
```diff
@@ -0,0 +1,388 @@
...
+end
+
+function interval(mesh, t)
+    if t == mesh[1]
+        return 1
+    elseif t == mesh[end]
+        return length(mesh) - 1
+    else
+        ind = findfirst(x -> x >= t, mesh)
+        i::Int64 = copy(ind)
```
ChrisRackauckas, line 36: why does this need a declaration?

ChrisRackauckas: oh the nothing case. Ehh that's kind of not nice but it works

src/adaptivity.jl
```diff
@@ -0,0 +1,388 @@
...
+    #TODO: Need users to manually specify, here, we set it as 3000 by default.
+    MxNsub = 3000
+
+    safety_factor = 1.3
+    rho = 1.0 # Set rho=1 means mesh distribution will take place everytime.
+    upper_new_mesh = 4.0
+    lower_new_mesh = 0.5
+    r1 = 0.0
+    r2 = 0.0
+    Nsub_star = 0
```
ChrisRackauckas, line 59: Surface things as kwargs in a follow up?

src/adaptivity.jl
```diff
@@ -0,0 +1,388 @@
...
+        h = mesh_current[i + 1] - mesh_current[i]
+        norm = abs(defect[i, idamax(defect[i, :])])
+        s_hat[i] = (norm / abstol)^(1.0 / (p + 1)) / h
+        if s_hat[i] * h > r1
+            r1 = s_hat[i] * h
+        end
+        r2 = r2 + s_hat[i] * h
+    end
+    r3 = r2 / n
+    n_predict::Int64 = round(Int, (safety_factor * r2) + 1)
```
ChrisRackauckas, line 74: Why `Int64` instead of `Int`?

src/adaptivity.jl
```diff
@@ -0,0 +1,388 @@
...
+    if abs((n_predict - n) / n) < 0.1
+        n_predict = round(Int, 1.1 * n)
+    end
+
+    if r1 <= rho * r3
+        Nsub_star::Int64 = 2 * n
```
ChrisRackauckas, line 80: what's the reason for this type declaration, and why Int64?

src/adaptivity.jl
```diff
@@ -0,0 +1,388 @@
...
+            Nsub_star = upper_new_mesh * n
+        end
+        if Nsub_star < lower_new_mesh * n
+            Nsub_star = lower_new_mesh * n
+        end
+        if Nsub_star > MxNsub
+            # Mesh redistribution fails
+            #println("New mesh would be too large")
+            info = ReturnCode.Failure
+            mesh_new = Nothing
```
ChrisRackauckas, line 100: Don't use the datatype, that won't specialize

src/adaptivity.jl
```diff
@@ -0,0 +1,388 @@
...
+            t = mesh_new[i + 2]
+            i += 1
+            integral = 0
+        else
+            integral += next_piece
+            t = mesh_current[k + 1]
+            k += 1
+        end
+    end
+    mesh_new[Nsub_star + 1] = mesh_current[n + 1]
```
ChrisRackauckas, line 142: This will fail on 32-bit OS with the coercion to Int64.

src/adaptivity.jl
```diff
@@ -0,0 +1,388 @@
...
+    sum = 0.0
+    for k in 1:n
+        sum += s_hat[k] * (mesh_current[k + 1] - mesh_current[k])
+    end
+    zeta = sum / Nsub_star
+    k::Int64 = 1
+    i::Int64 = 0
+    mesh_new[1] = mesh_current[1]
+    t = mesh_current[1]
+    integral::Float64 = 0.0
```
ChrisRackauckas, line 128: Why are these type declarations needed?

src/adaptivity.jl
```diff
@@ -0,0 +1,388 @@
...
+
+Divide the original subinterval into two equal length subinterval.
+"""
+function half_mesh(mesh_current::Vector)
+    n = length(mesh_current) - 1
+    mesh_new = zeros(Float64, 2 * n + 1)
+    mesh_new[1] = mesh_current[1]
+    for i in 1:n
+        mesh_new[2 * i + 1] = mesh_current[i + 1]
+        mesh_new[2 * i] = (mesh_current[i + 1] + mesh_current[i]) / 2.0
```
ChrisRackauckas, line 159:
```suggestion
        mesh_new[2 * i] = (mesh_current[i + 1] + mesh_current[i]) / 2
```

src/adaptivity.jl
```diff
@@ -0,0 +1,388 @@
...
+"""
+    half_mesh(mesh_current)
+
+The input mesh_current has length of n+1
+
+Divide the original subinterval into two equal length subinterval.
+"""
+function half_mesh(mesh_current::Vector)
+    n = length(mesh_current) - 1
+    mesh_new = zeros(Float64, 2 * n + 1)
```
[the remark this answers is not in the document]
ChrisRackauckas: In fact we shouldn't assume array here either, but make an issue for that.

src/adaptivity.jl
```diff
@@ -0,0 +1,388 @@
...
+    s, s_star, tau_star = TU.s, TU.s_star, TU.tau
+
+    # Initialization
+    defect = zeros(n, len)
+    #s, s_star, tau_star, x_star, v_star, c_star = setup_coeff(alg)
+
+    f_sample_1, f_sample_2 = zeros(Float64, len), zeros(Float64, len)
+    def_1, def_2 = zeros(Float64, len), zeros(Float64, len)
+    temp_1, temp_2 = zeros(Float64, len), zeros(Float64, len)
+    estimate_1, estimate_2 = zeros(Float64), zeros(Float64)
```
ChrisRackauckas, line 197: Assumes Float64

src/adaptivity.jl
```diff
@@ -0,0 +1,388 @@
...
+
+    # Evaluate at the first sample point
+    weights_1, weights_1_prime = interp_weights(tau_star, alg)
+    # Evaluate at the second sample point
+    weights_2, weights_2_prime = interp_weights(1.0 - tau_star, alg)
```
ChrisRackauckas, line 202:
```suggestion
    weights_2, weights_2_prime = interp_weights(1 - tau_star, alg)
```

etc. shouldn't assume float64

src/adaptivity.jl
```diff
@@ -0,0 +1,388 @@
...
+
+"""
+    interp_weights(tau, alg)
+
+interp_weights: solver-specified interpolation weights and its first derivative
+"""
+function interp_weights(tau, alg::Union{GeneralMIRK, MIRK})
+    if alg_order(alg) == 4
+        t2 = tau * tau
+        tm1 = tau - 1.0
```
ChrisRackauckas, line 325:
```suggestion
        tm1 = tau - 1
```

ChrisRackauckas: Etc., shouldn't assume Float64

src/cache.jl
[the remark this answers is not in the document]
ChrisRackauckas: It's mutating the values but not the cache itself

src/adaptivity.jl
ErikQQY: I think while we can know the type of ```tau_star```, why don't we directly use Float to avoid abundant type conversion?

src/adaptivity.jl
```diff
@@ -0,0 +1,388 @@
...
+        wp = [-tm1 * t4m3 * t2m1 / 3.0,
+            tau * t2m1 * t4m3,
+            4.0 * tau * t4m3 * tm1,
+            -32.0 * tau * t2m1 * tm1 / 3.0]
+
+        return w, wp
+
+    elseif alg_order(alg) == 6
+        w = [
+            tau - 28607 / 7434 * tau^2 - 166210 / 33453 * tau^3 + 334780 / 11151 * tau^4 -
```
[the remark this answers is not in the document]
ErikQQY: Like what we did in OrdinaryDiffEq.jl?

ChrisRackauckas:
Needs to rebase for the MIRK5, then I think we can merge and continue some of this in a follow up.

ErikQQY:
@ChrisRackauckas bump

ChrisRackauckas:
Open issues for the points that weren't addressed, but let's go for it. 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉 🎉

PR #90, opened 2023-05-30
Purge NLsolve and introduce NonlinearSolve.jl

avik-pal:


avik-pal:
How does the `test_convergence` function work? I was comparing the numerical solutions to the analytical solutions and both seem to match but the tests fail :(

cc @ChrisRackauckas

ChrisRackauckas (approved):

avik-pal:
testing

@ChrisRackauckas this is how the solutions look like

ChrisRackauckas:
No I mean plot the convergence simulation. Just use the plot recipe on it.

avik-pal:
testing

ChrisRackauckas:
just get rid of the last dt.

avik-pal:
It still seems to fail

ChrisRackauckas:
I meant the smallest dt, where the bump is.

avik-pal:
Not sure about the `orbital.jl` tests, they converge on my local machine

YingboMa (approved):

avik-pal:
I don't have commit access to the repo; someone else needs to merge!

YingboMa:
We still cannot merge it before fixing the orbit example though

avik-pal:
This is now ready to be merged!

avik-pal:
Bump @ChrisRackauckas /@YingboMa

src/algorithms.jl
```diff
@@ -1,16 +1,17 @@
+const DEFAULT_NLSOLVE_SHOOTING = TrustRegion(; autodiff = Val(false))
+const DEFAULT_NLSOLVE_MIRK = NewtonRaphson(; autodiff = Val(false))
```
ChrisRackauckas, line 2: why autodiff false?

ChrisRackauckas:
Just the one question on autodiff

src/algorithms.jl
avik-pal: I kept it false to ensure closer behavior to the NLsolve default we had previously.

ChrisRackauckas: With it being a breaking change, it would be best to get it in line with what we do elsewhere and default to forward mode AD

ChrisRackauckas, line 2:
```suggestion
const DEFAULT_NLSOLVE_SHOOTING = TrustRegion(; autodiff = Val(true))
const DEFAULT_NLSOLVE_MIRK = NewtonRaphson(; autodiff = Val(true))
```

PR #91, opened 2023-06-23
Fix stage in collocation

ErikQQY:
When we are calculating the subcomponents of the residual function

$$
\Phi_i=y_{i+1}-y_{i}-h\sum_{r=1}^sb_{r}K_{r}
$$

we are actually iterating from 1 to $s$ instead of the algorithm order.


ChrisRackauckas (approved):

ChrisRackauckas:
Yes the order and number of stages may not be the same.

PR #96, opened 2023-08-03
Multiple Fixes for Type Stability & Performance

avik-pal:
# Stuff Improved

* Makes the code type stable
* Float64 uses are mostly removed
* FiniteDiff was not using sparsity pattern before which has now been fixed!

# Performance

For the first problem in MIRK Convergence tests

```julia
### New Version
julia> @benchmark sol = solve(prob, GeneralMIRK5(), dt = 0.2)
BenchmarkTools.Trial: 9349 samples with 1 evaluation.
 Range (min … max):  301.631 μs … 43.935 ms  ┊ GC (min … max):  0.00% … 98.77%
 Time  (median):     341.470 μs              ┊ GC (median):     0.00%
 Time  (mean ± σ):   530.875 μs ±  2.325 ms  ┊ GC (mean ± σ):  29.84% ±  6.74%

  ▄▇██▇▆▆▅▄▄▃▃▃▃▃▂▂▂▁▁▁▁▁▁▁▁▁▁                                 ▂
  █████████████████████████████▇▇▇▇▇▇▅▆▆▅▅▅▅▆▅▅▅▄▅▅▄▅▄▃▄▄▅▄▅▃▆ █
  302 μs        Histogram: log(frequency) by time       865 μs <

 Memory estimate: 796.89 KiB, allocs estimate: 9147.

### Old Version
julia> @benchmark solve($prob, $GeneralMIRK5(); dt=0.2)
BenchmarkTools.Trial: 1431 samples with 1 evaluation.
 Range (min … max):  2.067 ms … 30.413 ms  ┊ GC (min … max):  0.00% … 86.92%
 Time  (median):     2.469 ms              ┊ GC (median):     0.00%
 Time  (mean ± σ):   3.486 ms ±  3.236 ms  ┊ GC (mean ± σ):  19.84% ± 17.74%

  █▆▄▃▂▅▅▃                                                    
  █████████▆▄▇▄▄▅▄▄▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▄▁▁▁▁▁▁▁▁▄▄▄▄▆▆▆▄▇▇▇▆▅▄ █
  2.07 ms      Histogram: log(frequency) by time     17.5 ms <

 Memory estimate: 5.08 MiB, allocs estimate: 66033.
```

# Testing Problems

### New

new

### Old

old

The convergence tests seem to fail despite the final error being lower. Not sure how to handle that. :sweat:


avik-pal:
I seem to have figured out the problem, using the sparsity code makes the jacobian `J[1, 1]` incorrect.

@ChrisRackauckas can you take a look at how I am specifying the sparsity pattern and if I am doing something obviously stupid?

avik-pal:
I am a bit lost on how to fix the order tests :sweat:

ErikQQY:
I tested the MIRK5 method and found that if we increase the order of ```abstol```, the order tests could pass.

```julia
@time sim = test_convergence(dts1, prob, GeneralMIRK5(); abstol = 10^(-5.5), reltol = 1e-4);
```

ChrisRackauckas:
Increase? That looks like a decrease. I assume that's the abstol of the nonlinear solver, since adaptivity is turned off for the convergence tests? If that's all the case, then it would make sense to simply always have the tolerance as low as possible, like 1e-12, so that the convergence of the nonlinear solver does not interfere with the convergence test of the tableau.

avik-pal:
testing

I don't think the convergence thing is doing the right thing. This is the plot for the reduced tolerance GeneralMIRK5. For MIRK6, the errors were `0` for some abstol/reltol making the order estimate as NaN.

avik-pal:
I turn off adaptivity while testing for convergence.

ChrisRackauckas:
The convergence testing function should hardcode the adaptive=false thing, can you upstream that?

avik-pal:
It seems adaptive was false by default, but that option did nothing for BoundaryValueDiffEq. I have added the early termination if adaptive is false.

avik-pal:
@ChrisRackauckas I think this is ready to go! I have quite a few performance fixes in a followup PR. I believe having them together would be extremely messy for review purposes.

PR #102, opened 2023-08-19
Fix initial guess interface

### LINKED ISSUE #27: Improve initial guess interface
`BVProblem` and `TwoPointBVProblem` should accept a function `initialGuess( t )` and/or a solution structure in order to reuse solutions as initial guess for continuation.
The documentation on this should be improved/ updated accordingly.

### LINKED ISSUE #30: Dimension mismatch in solve() for MIRK methods providing an initial solution
`BVPSolve` is allocating more memory than neccessary when provided a initial guess. It is basically allocating a `n` times `n` Vector of Vectors, where `n` is the number of timepoints.

Additionally i get a `DimensionMismatch` in https://github.com/JuliaDiffEq/BoundaryValueDiffEq.jl/blob/d6b2bcace6c743250366bfea67d2366c4d79d018/src/solve.jl#L55.

It would be good if the dimensions of the initial solution where checked to have the right dimensions and throw a corresponding error if not.

That could be done when the `BVProblem` gets defined. My testing code can be found [here](https://gitlab.gwdg.de/PhD/shapeJulia/blob/master/mwe.jl).

ErikQQY:
Fix #27 
Fix #30

The PR fixes the initial guess interface of the MIRK method. It is just working but still needs further optimization to avoid potential bugs.


src/solve.jl
```diff
@@ -31,8 +33,9 @@ end
 function DiffEqBase.__solve(prob::BVProblem, alg::AbstractMIRK; dt = 0.0, abstol = 1e-3,
     adaptive::Bool = true, kwargs...)
     dt ≤ 0 && throw(ArgumentError("dt must be positive"))
-    T = eltype(prob.u0)
-    n = Int(cld((prob.tspan[2] - prob.tspan[1]), dt))
+    T = isa(prob.u0[1], Vector) ? eltype(prob.u0[1]) : eltype(prob.u0)
+    n = isa(prob.u0[1], Vector) ? (length(prob.u0) - 1) :
```
ErikQQY, line 37: Judging if the provided ```u0``` is an initial guess or the initial condition in this way is pretty awkward, do we have any other solutions to check the type of ```u0```?

ErikQQY:
It seems the CI failing is caused by Julia 1.6 version:

CI

src/solve.jl
ChrisRackauckas: I don't think we have a better way.

ChrisRackauckas:
I think we can just error on that for now with a specific error that says this feature is unsupported on Julia < v1.9 and move on.

avik-pal:
```julia
@static if VERSION ≥ v"1.7" # Check when the function was introduced
    amax(x; kwargs...) = first(findmax(abs, x; kwargs...))
else
    amax(x; kwargs...) = first(findmax(abs.(x); kwargs...))
end
```

src/mirk_tableaus.jl
```diff
@@ -115,8 +115,8 @@ function constructMIRK6(S::BVPSystem)
 
     # Interpolant tableau
     s_star = 9
-    c_star = [7 // 16, 1 // 8, 9 // 16, 3 // 8]
-    v_star = [7 // 16, 1 // 8, 9 // 16, 3 // 8]
+    c_star = [7 // 16, 3 // 8, 9 // 16, 1 // 8]
+    v_star = [7 // 16, 3 // 8, 9 // 16, 1 // 8]
```
ChrisRackauckas, line 119: Is a convergence test checking htis?

ErikQQY: The interpolation coefficients don't affect the convergence test, but they did affect the ensemble test. I found the previous interpolation coefficients in MIRK6 are wrong, so I directly changed them in this PR.

```fortran
c_star(1) = 7.0d0/16.0d0
c_star(2) = 3.0d0/8.0d0
c_star(3) = 9.0d0/16.0d0
c_star(4) = 1.0d0/8.0d0

v_star(1) = 7.0d0/16.0d0
v_star(2) = 3.0d0/8.0d0
v_star(3) = 9.0d0/16.0d0
v_star(4) = 1.0d0/8.0d0
```

ChrisRackauckas: Let's do a follow up with some convergence or regression test on the interpolation. The interpolation coefficients should be used as the post solution interpolation, so I think that's the first priority: make `sol(t)` work at higher order using this continuous definition (and save the `k`'s to the solution). Once that's the case, we can add a dense convergence test which will use non-grid values to test the convergence, and that should then test the accuracy of the interpolation.

PR #103, opened 2023-08-30
Update Project.toml

ErikQQY:
We need a new release for the benckmarks


ChrisRackauckas:
Actually, this needs to be a v4.0 since GeneralMIRK was removed.

PR #98, opened 2023-08-08
Faster Jacobian via Split backends

avik-pal:
## Goals

- [x] ~Fast Code Path for TwoPointBVProblem (Banded Matrix so we can do complete sparse diff)~
  - [x] ~Rewrite the loss function to reorder the outputs correctly.~
- [x] Split Mode
  - [x] Sparse Diff for the Banded Part
  - [x] Dense Diff for the `bc!` Part
    * We won't be able to use Zygote since `bc!` assumes in-place operations
    * The Jacobian has `inputs > outputs` so we want to use ReverseMode ideally. `Enzyme` might be a good choice. For fallback, we can just use `ReverseDiff`.
- [x] Sparsity Detection for `bc!` and if sparse, then use SparseDiff here as well
  - [x] Requires Sparse ReverseMode (Not implemented upstream yet)

PR #106, opened 2023-09-01
Towards GPU Compatible BVP Solvers

avik-pal:
* Rolls back the matrix / Array{3} internal representations
* Returns to a vector of arrays representation but performs all operations in-place if possible  
    * Previous implementation was vector of arrays but still reallocated everything
* GPUs should work with the current version (still to be tested)
* Needs a merge of #98 first
* Overall allocations are significantly down (tested against master)

## TODOs

- [ ] Does `f` on GPU work?
    * Really bad performance currently -- BandedMatrix doesn't quite work
    * coloring fails for CuSparse
- [x] Can `u` be anything other than a vector?


.JuliaFormatter.toml
```diff
@@ -1 +1,2 @@
-style = "sciml"
\ No newline at end of file
+style = "sciml"
+annotate_untyped_fields_with_any = false
```
ChrisRackauckas, line 2: isn't this already in the sciml format?

ChrisRackauckas: It doesn't do that 😅

avik-pal: https://github.com/domluna/JuliaFormatter.jl/blob/c72d049210a06e2a404df51998051f3dbb0a123b/src/styles/sciml/pretty.jl#L25

It seems okay to have it as true by default, but if we use something like ConcreteStructs (which IMO we should be using everywhere), then it becomes a problem.

src/adaptivity.jl
```diff
@@ -115,149 +107,153 @@ function redistribute(mesh, Nsub_star::Int, ŝ::AbstractVector{T}, dt) where {T
...
-The input mesh has length of n+1
+    half_mesh!(mesh, mesh_dt)
+    half_mesh!(cache::MIRKCache)
 
-Divide the original subinterval into two equal length subinterval.
+The input mesh has length of `n + 1`. Divide the original subinterval into two equal length
+subinterval. The `mesh` and `mesh_dt` are modified in place.
 """
-@views function half_mesh(mesh::AbstractVector{T}) where {T}
+function half_mesh!(mesh::Vector{T}, mesh_dt::Vector{T}) where {T}
```
ChrisRackauckas, line 123: only Vector?

avik-pal: Mesh is controlled internally by us. Irrespective of what the input types are we always construct a Vector. We can relax it, but I am not sure how many types declare stuff like `resize!`

ChrisRackauckas: Won't the mesh need to be a GPU vector for the GPU stuff to work?

avik-pal: Not currently. But again, the code right now is very slow on GPU.

The main issue is we iterate over the mesh to redistribute the new points, so we need to assume fast scalar indexing. We will have to figure out a way to batch the entire computation (which technically is trivial to write a kernel for).

PR #109, opened 2023-09-12
Fast path for Two Point BVPs

avik-pal:
- [x] Tests from ODEInterface.jl


ChrisRackauckas:
Looks reasonable but tests fail?

ext/BVPODEInterfaceExt.jl
```diff
@@ -0,0 +1,95 @@
+module BVPODEInterfaceExt
```
ChrisRackauckas, line 1: let's use the full name 😅

avik-pal:
This will have to wait till we release the NonlinearSolve 2.0. The released versions can't handle residual prototype it seems so they wont work for two point BVPs.

ext/BoundaryValueDiffEqODEInterfaceExt.jl
```diff
@@ -0,0 +1,113 @@
...
+            Termination since multiple shooting condition or
+            condition of Jacobian is too bad for `sol_method=1`"""
+        elseif retcode == -8
+            @error "Condensing algorithm for linear block system fails, try `sol_method=1`"
+        elseif retcode == -9
+            @error "Sparse linear solver failed"
+        elseif retcode == -10
+            @error "Real or integer work-space exhausted"
+        elseif retcode == -11
+            @error "Rank reduction failed - resulting rank is zero"
```
ChrisRackauckas, line 105:
```suggestion
            @warn "Integrator failed to complete the trajectory"
        elseif retcode == -4
            @warn "Gauss Newton method failed to converge"
        elseif retcode == -5
            @warn "Given initial values inconsistent with separable linear bc"
        elseif retcode == -6
            @warn """Iterative refinement faild to converge for `sol_method=0`
            Termination since multiple shooting condition or
            condition of Jacobian is too bad for `sol_method=1`"""
        elseif retcode == -8
            @warn "Condensing algorithm for linear block system fails, try `sol_method=1`"
        elseif retcode == -9
            @warn "Sparse linear solver failed"
        elseif retcode == -10
            @warn "Real or integer work-space exhausted"
        elseif retcode == -11
            @warn "Rank reduction failed - resulting rank is zero"
```

Never block.

src/algorithms.jl
```diff
@@ -1,5 +1,5 @@
-const DEFAULT_NLSOLVE_SHOOTING = TrustRegion(; autodiff = Val(true))
-const DEFAULT_NLSOLVE_MIRK = NewtonRaphson(; autodiff = Val(true))
+const DEFAULT_NLSOLVE_SHOOTING = NewtonRaphson()
+const DEFAULT_NLSOLVE_MIRK = NewtonRaphson()
```
ChrisRackauckas, line 2: Not so sure about that.

ChrisRackauckas:
Is there a reason colnew isn't supported here? Have you checked its interface to see that it'll be fine?

avik-pal:
`colnew` needs some manipulation on our end to match the interface. But we don't need to add anything additional on SciMLBase side.

src/algorithms.jl
avik-pal: This change to NewtonRaphson was before the TR bugs were fixed. We can revert this part. The `autodiff = true` is redundant wince that is the current default.

PR #110, opened 2023-09-13
Porting over Multiple Shooting and Single Shooting from NeuralBVP

avik-pal:
~**Don't merge before https://github.com/SciML/BoundaryValueDiffEq.jl/pull/109**~

* Single shooting now takes `odesolve_kwargs` and `nlsolve_kwargs` to allow different kwargs. repetition in `kwargs` overrides these choices
* Single shooting can now handle non-vector inputs.
* Working Multiple Shooting Code with Grid Coarsening

## TODOs

- [x] Port tests from NeuralBVP.jl
- [x] ~Translate NeuralBVP.jl to use BoundaryValueDiffEq.jl~ (Upstream issue)
- [x] Incorporate using Initial Guess in Shooting Codes
- [x] Use Jacobian sparsity to speed up Newton Steps!
  - [x] Sparse Jacobian Prototype Constructor
  - [x] Hardcoded version using ForwardDiff
    - [x] Threading causes Segfaults
  - [x] Use Coloring
  - [x] Generalize
- [x] Two Point BVP Fast Path
- [x] ~Test BVP with NLS~ (Will do it in a separate PR)


test/shooting/shooting_tests.jl
```diff
@@ -0,0 +1,272 @@
...
+    for solver in SOLVERS
+        sol = solve(bvp4, solver; abstol = 1e-6, reltol = 1e-6)
+        @test SciMLBase.successful_retcode(sol)
+        resid_f = reduce(vcat, bc2((sol(tspan[1]), sol(tspan[2])), nothing))
+        @test norm(resid_f) < 1e-6
+    end
+end
+
+@testset "Shooting with Complex Values" begin
+    # Test for complex values
```
ErikQQY, line 92:
```suggestion
    # Test for complex values
        function f1!(du, u, p, t)
        du[1] = u[2]
        du[2] = -u[1]
        return nothing
    end

    function bc1!(resid, sol, p, t)
        t₀, t₁ = first(t), last(t)
        resid[1] = sol(t₀)[1]
        resid[2] = sol(t₁)[1] - 1
        return nothing
    end
```

avik-pal:
This PR got more bloated than I anticipated. But it should be done now. Takeaways:

* For `MS` using the two-point form is significantly faster
* For the Orbital problem, we need to use TrustRegion for proper convergence
* Solving an Ensemble Problem for the MS gives the most gains.
* There was a bug in the interpolation code, which has been patched

TODOs for later:

* MS can't handle Non-Vector like Inputs. I will deal with it together with the handling for matrix inputs to handle Neural Networks better!

ChrisRackauckas (approved):

ChrisRackauckas:
🎉 It was a bit hard to read since there were multiple things mixed in there, but generally fine so merging and will deal with anything else in the future.

PR #112, opened 2023-09-19
Add interpolations for MIRK methods

### LINKED ISSUE #111: Interoplation for MIRK methods
We need to implement interpolations for MIRK methods so that we can approximate the solution between mesh points.

ErikQQY:
Fix #111

This PR adds interpolations for MIRK methods

Tested on example problem from README:

image


ChrisRackauckas:
Can you add a convergence with something like a mid time condition in order test the correctness here?

avik-pal:
This only adds interpolation to the solution object, right? We need the intermediate objects passed to the `bc` to be interpolation-friendly as well.

ChrisRackauckas:
Yes, that `sol` object is what you should just be passing to the bc anyways, just then updating its coefficient vectors as you go along.

avik-pal:
This would also be a breaking change, right? We can up the SciMLBase compat to 2.0 and do that together with https://github.com/SciML/BoundaryValueDiffEq.jl/pull/109 & https://github.com/SciML/SciMLBase.jl/pull/477

ChrisRackauckas:
Why would this be a breaking change?

avik-pal:
That changes the input specification to `bc`. The current specification is a vector of arrays which now changes to a solution similar to shooting.

ChrisRackauckas:
It was supposed to always be the solution. The solution type acts as a vector of arrays so the actions are the same, just extended though?

src/adaptivity.jl
```diff
@@ -253,7 +253,7 @@ end
         w′[(stage + 1):s_star],
         true,
         true)
-    z .= z .* dt .+ cache.y₀[i]
+    z .= z .* dt[1] .+ cache.y₀[i]
```
ChrisRackauckas, line 256: this `1` seems odd?

PR #113, opened 2023-09-21
WIP: Add Lobatto IIIa-b and Radau II tableaus

axla-io:
This PR adds Lobatto and Radau II RK methods to calculate the collocation loss for a BVP

Checklist:
- [x] Add Lobatto IIIa tables
- [x] Add Lobatto IIIb tables
- [x] Add Radau II tables
- [x] Implement RK collocation
- [ ] Adaptivity (Interpolated tables)
- [ ] Sparse Jacobian for RK collocation
- [ ] Add tests
- [ ] Add docstrings

### Questions / comments
1. I renamed `MIRKCache` into an `RKCache`. I also added a new `RKTableau` struct for holding standard RK tableaus. Then I added a new collocation function that gets dispatched on the type of Tableau that is in the cache. Idk if this is the best way to set up dispatching, but it is pretty concise.
2. For a simple pendulum problem, the Lobatto IIIb collocation seems to match the MIRK results when it converges, but frequently runs into NaN solutions. I've tested the tableau values against `RungeKutta.jl`. The collocation implementation itself is based on the one for MIRKs with a small adaptation to make it work with a full Butcher's tableau. The NaN solution problem is not mitigated by decreasing dt, as far as I've checked (in the range 0.1 to 0.001).
3. Assuming that we can use the same machinery for adaptive stepping as for MIRK, the interpolation tables for the RK methods need to be derived. How can I do that?
4. Sparse Jacobian as previously implemented for MIRKs doesn't work. The method needs to be modified. Currently I'm using an autodiff Jacobian (the default in `NonLinearSolve`) so solves are slow.


axla-io:
image
Example of a good run, where only the five stage version gets NaNs. I haven't been able to get the 5 stage version of Lobatto IIIb to converge.

avik-pal:
@axla-io  can you rebase and check if the sparse jacobian problem still exists?

axla-io:
Oh I'll take a look tomorrow!

axla-io:
@avik-pal It still gets singular exception :(

PR #120, opened 2023-10-12
Minor Updates to Multiple Shooting

avik-pal:


src/solve/multiple_shooting.jl
```diff
@@ -236,7 +231,7 @@ function __solve(prob::BVProblem, _alg::MultipleShooting; odesolve_kwargs = (;),
                 nodes); resid_prototype, jac = jac_fn, jac_prototype)
         nlprob = NonlinearProblem(loss_function!, u_at_nodes, prob.p)
         sol_nlsolve = __solve(nlprob, alg.nlsolve; nlsolve_kwargs..., verbose, kwargs...)
-        # u_at_nodes = sol_nlsolve.u
+        u_at_nodes = sol_nlsolve.u::typeof(u0)
```
ChrisRackauckas, line 234: why is the type assert needed?

avik-pal: I tried figuring out where the exact type instability is coming from, but Cthulhu also breaks down...

PR #124, opened 2023-10-16
Setup for using Banded Structures in BVP

avik-pal:
- [x] Upstream Banded QR Piracy https://github.com/SciML/LinearSolve.jl/pull/393
- [x] Adjoint / Transpose Banded Matrices proper support https://github.com/JuliaArrays/ArrayInterface.jl/pull/422
- [x] Support for MIRK Methods
- [x] Support for Multiple Shooting
    - [x] Implement Reordering for MS
- [x] Change defaults for linear solve (banded default selection doesn't work) https://github.com/SciML/LinearSolve.jl/pull/393
- [ ] CUDA / AMDGPU scalar indexing patch for Banded Matrices
- [x] Component Arrays


src/BoundaryValueDiffEq.jl
```diff
@@ -16,6 +16,14 @@ import SparseDiffTools: AbstractSparseADType
 import TruncatedStacktraces: @truncate_stacktrace
 import UnPack: @unpack
 
+# FIXME: Remove
+function NonlinearSolve.LinearSolve.init_cacheval(alg::NonlinearSolve.LinearSolve.KrylovJL,
+    A::BandedMatrix, b::ArrayPartition, u, Pl, Pr, maxiters::Int, abstol, reltol,
+    verbose::Bool, ::NonlinearSolve.LinearSolve.OperatorAssumptions)
+    return nothing
+end
+# FIXME: Remove
```
ChrisRackauckas, line 25: Upstream before merging

avik-pal:
@ChrisRackauckas should we go ahead with this? If using CUDA / AMDGPU then we throw a fairly obvious error. Supporting it properly needs some custom cuda kernels (the state of CUSPARSE seems weird with most operations sending the arrays back to CPU)

ChrisRackauckas:
> Supporting it properly needs some custom cuda kernels (the state of CUSPARSE seems weird with most operations sending the arrays back to CPU)

Yes. GPU needs a lot more. It needs custom kernel generation for how to construct the Jacobian, and then special CUSPARSE routines where we use the ilu0 (which is missing ldiv!'s). It needs more work, let's do that part separately.

avik-pal:
The CUDA part will unfortunately be a massive blocker if we want to Neural BVPs :disappointed:

ChrisRackauckas:
At least MultipleShooting is ready for CUDA?

avik-pal:
Need to test, but most likely no. It also specializes heavily on sparsity patterns

avik-pal:
For the purposes of Neural BVP if we avoid passing in the jacobian function and just use GMRES we should be good to go (I think)

ChrisRackauckas:
Yes that should be the case. Jacobian-free should be fine. Anything with a Jacobian involved will take more work (but be better overall). But that's why I had it as a separate project 😅

PR #125, opened 2023-10-17
[WIP] Move from ArrayPartition to ComponentArrays

avik-pal:
- [ ] MIRK
  - [x] Get basic version working
  - [ ] Rework to pass a vector prototype into the nlfunction
- [x] Shooting
- [ ] Multiple Shooting

Array Partition seems to allocate too much. I could shave off quite a lot of time for small problems with this shift!


ChrisRackauckas:
Yes they will allocate when doing linear algebra because they are non-contiguous. Also you want to generally avoid their linear indexing.

avik-pal:
> Yes they will allocate when doing linear algebra because they are non-contiguous. Also you want to generally avoid their linear indexing.

It makes sense to make the move then, since we are going to be doing a lot of matvecs in the nonlinear solver

avik-pal:
Pulling this into the Banded Matrix PR since I need it to test GPU

PR #126, opened 2023-10-20
Build ODESolution type with MIRK Interpolation

difeizhang:
- [ ] build ODESolution type for ::StandBVProblem
  - [x] pass original tests
  - [ ] add new MIRK tests with interpolation interface of boundary conditions
  - [ ] pass new tests


difeizhang:
new PR.already using JuliaFormatter. 
1. build BVProblem solution as ODESolution type with MIRK interpolation. allow set boundary conditions at any time.
2. add conditional inside loss_bc, based on whether the problem type is two point or multi point
3. change pendulum boundary condition in the mirk test
@ChrisRackauckas @avik-pal

avik-pal:
@difeizhang tests are broken

difeizhang:
I am looking into it now. Seems broken for high order MIRK with error SingularException0.

avik-pal:
@difeizhang how did the indexing problem you mentioned get fixed?

difeizhang:
> @difeizhang how did the indexing problem you mentioned get fixed?

@avik-pal I partially fix the indexing issue, but haven't pushed it. Now I just set calculate_error = false to pass original tests.
When first time building solution and doing interpolation, the error occurs in the `adaptivity.jl`. 
`function sum_stages!(z, cache::MIRKCache, w, i::Int; u = cache.y₀, dt = cache.mesh_dt[i])`
`    @unpack M, stage, mesh, k_discrete, k_interp, mesh_dt = cache`
`    @unpack s_star = cache.ITU`
`    z .= zero(z)` 
`    __maybe_matmul!(z, k_discrete[i].du[:, 1:stage], w[1:stage])`
`    __maybe_matmul!(z, k_interp[i][:, 1:(s_star - stage)],  w[(stage + 1):s_star], true, true)  <--- `
`    z .= z .* dt .+ u[i]`
`    return z`
`end`
The k_interp[i] is empty (0x0 matrix) and cannot be indexing. So I add conditional `if !isempty(k_interp[i])` for it.

I am still debugging it. Now it supports both array and interpolation interface. But the solution is not accurate now. Seems the Jacobian is changed. I will push it when everything is done correctly and can pass the new tests.

difeizhang:
I'll do the way you suggest to change the interp_setup

avik-pal:
@difeizhang what is the status on this?

Issue #128, opened 2023-11-01
Support Initial Guess Function


avik-pal:
As pointed out in https://github.com/SciML/BoundaryValueDiffEq.jl/issues/117 this is not currently supported.

- [x] MIRK
- [x] Multiple Shooting
- [x] Shooting (should just default to picking the initial condition)

Issue #117, opened 2023-10-09
Very long compilation time for solve()


cpf-work:
I've upgraded to v5.1 today and found very long compilation times when trying the simple pendulum example.

My base environment:
  [6e4b80f9] BenchmarkTools v1.3.2
  [7073ff75] IJulia v1.24.2
  [295af30f] Revise v3.5.6
  [0f1e0344] WebIO v0.8.21
  [de0858da] Printf
  
I created a minimal environment for testing:
  [764a87c0] BoundaryValueDiffEq v5.1.0
  [91a5bcdd] Plots v1.39.0

I'm using Julia 1.9.3.

I run ["Example 1: Simple pendulum"](https://docs.sciml.ai/DiffEqDocs/stable/tutorials/bvp_example/#Example-1:-Simple-Pendulum) from the docs and find

```@time sol1 = solve(bvp1, MIRK4(), dt = 0.05)```
225.577779 seconds (289.71 M allocations: 60.160 GiB, 6.70% gc time, 99.98% compilation time)

Second execution is fast:
```@time sol1 = solve(bvp1, MIRK4(), dt = 0.05)```
0.001343 seconds (6.28 k allocations: 879.165 KiB)

I'm quite confident this did not happend when I played around 1-2 weeks ago with an earlier version. I tried reverting to v5.0.0, but I'm not experienced in doing stuff like this and obtained an error message upon `solve()`, so I can't tell whether this also happens for me with v5.0.0 now.

I managed to revert to v4.0.1 
⌃ [764a87c0] BoundaryValueDiffEq v4.0.1

And I get more reasonable compilation times:
```@time sol1 = solve(bvp1, MIRK4(), dt = 0.05)```
 28.803127 seconds (42.95 M allocations: 6.039 GiB, 6.34% gc time, 99.76% compilation time)

mateuszbaran:
:+1:

It's even worse in Manifolds.jl (https://github.com/JuliaManifolds/Manifolds.jl/pull/657):

13210.001646 seconds (24.48 G allocations: 7.394 TiB, 9.48% gc time)

vs BoundaryValueDiffEq v4.0.1:

142.380662 seconds (234.62 M allocations: 17.268 GiB, 4.31% gc time, 82.72% compilation time: 1% of which was recompilation)

ChrisRackauckas:
Does it go away with Julia v1.10?

cpf-work:
[...]

PR #131, opened 2023-11-02
Reuse ODE Solver Cache

avik-pal:
Fixes https://github.com/SciML/BoundaryValueDiffEq.jl/issues/26

- [x] Shooting
- [x] Multiple Shooting

Locally I tested Shooting improves by ~10%.

Need to do a breaking change since we drop the NonlinearSolver autodiff now!

PR #133, opened 2023-11-04
Support for u0 initial guess function

avik-pal:
Fixes https://github.com/SciML/BoundaryValueDiffEq.jl/issues/128

## TODOs

- [x] Upstream `get_concrete_u0` patch so that `u0` as a function is propagated unlike IVPs
- [x] Add tests (use the one in https://github.com/SciML/BoundaryValueDiffEq.jl/issues/117#issuecomment-1780981510)


avik-pal:
I will fix the piracy once https://github.com/SciML/DiffEqBase.jl/pull/959 is merged and released.

avik-pal:
I want to get this in before https://github.com/SciML/BoundaryValueDiffEq.jl/pull/131, since that one is major release.

src/utils.jl
```diff
@@ -139,17 +139,38 @@ function __extract_problem_details(prob, u0::AbstractArray; dt = 0.0,
...
+    u0 = __initial_guess(f, prob.p, prob.tspan[1])
+    t₀, t₁ = prob.tspan
+    return Val(true), eltype(u0), length(u0), Int(cld(t₁ - t₀, dt)), u0
 end
 
-__initial_state_from_prob(prob::BVProblem, mesh) = __initial_state_from_prob(prob.u0, mesh)
-__initial_state_from_prob(u0::AbstractArray, mesh) = [copy(vec(u0)) for _ in mesh]
-function __initial_state_from_prob(u0::AbstractVector{<:AbstractVector}, _)
+function __initial_guess(f::F, p::P, t::T) where {F, P, T}
+    if static_hasmethod(f, Tuple{P, T})
```
ChrisRackauckas, line 152: note that in v1.10 this isn't needed, and it can be incorrect.

src/utils.jl
```diff
@@ -139,17 +139,38 @@ function __extract_problem_details(prob, u0::AbstractArray; dt = 0.0,
...
+        return f(p, t)
+    elseif static_hasmethod(f, Tuple{T})
+        return f(t)
+    else
+        throw(ArgumentError("`initial_guess` must be a function of the form `f(p, t)` or \
+                             `f(t)`"))
```
ChrisRackauckas, line 158: We should just simplify it to be just one form.

avik-pal: We can only allow `f(t)` according to the SciMLBase specification but I think we almost need the parameters

avik-pal: Ideally we just support `f(p, t)`. But that is a breaking change following the SciMLBase docs

ChrisRackauckas: throw an warning that f(t) is deprecated. Might as well start pushing in the right direction.

avik-pal: So let it work right now with the `static_hasmethod` but just throw a warning, correct?

ChrisRackauckas: yes

avik-pal: Done

Issue #123, opened 2023-10-13
Autoselect Correct Solver for Nonlinear Least Squares


avik-pal:
Currently, we fix the solver at the time of algorithm creation. This means for a `NLS` problem our defaults are incorrect.

Easy solution:
* Keep `nlsolve` as nothing
* Run a `concretize` step
   * If not `nothing` then return the solver
   * If `nothing` choose the solver based on the generated problem

avik-pal:
I tried this but it completely kills type stability. Which is not unexpected since the solver choice is dependent on runtime parameters.

avik-pal:
NonlinearSolve.jl has a `nothing` dispatch for `NonlinearProblem`. All we need to do now is add one for `NonlinearLeastSquaresProblem`. Once that is done, we can set the default `nlsolve` to be `nothing`

PR #137, opened 2023-11-06
Use PolyAlgorithms for Nonlinear Solves

avik-pal:
Fixes https://github.com/SciML/BoundaryValueDiffEq.jl/issues/123


avik-pal:
Now we don't even need to do anything different for NLLS, if `size(du) != size(u)` it will automatically construct an NLLSProblem and NonlinearSolve will dispatch to the correct solver!

PR #140, opened 2023-11-07
Fix solution type from BVPM2 and BVPSOL

ErikQQY:
Now the solution type from ODEInterface.jl is compatible with the others

### DIFF
--- a/ext/BoundaryValueDiffEqODEInterfaceExt.jl
+++ b/ext/BoundaryValueDiffEqODEInterfaceExt.jl
@@ -53,8 +53,9 @@ function __solve(prob::BVProblem, alg::BVPM2; dt = 0.0, reltol = 1e-3, kwargs...
     retcode = retcode ≥ 0 ? ReturnCode.Success : ReturnCode.Failure
 
     x_mesh = bvpm2_get_x(sol)
+    evalsol = evalSolution(sol, x_mesh)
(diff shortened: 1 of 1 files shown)


ext/BoundaryValueDiffEqODEInterfaceExt.jl
```diff
@@ -53,8 +53,9 @@ function __solve(prob::BVProblem, alg::BVPM2; dt = 0.0, reltol = 1e-3, kwargs...
     retcode = retcode ≥ 0 ? ReturnCode.Success : ReturnCode.Failure
 
     x_mesh = bvpm2_get_x(sol)
+    evalsol = evalSolution(sol, x_mesh)
     sol_final = DiffEqBase.build_solution(prob, alg, x_mesh,
-        eachcol(evalSolution(sol, x_mesh)); retcode, stats)
+        Vector{eltype(evalsol)}[eachcol(evalsol)...]; retcode, stats)
```
ChrisRackauckas, line 58: Don't splat.

ChrisRackauckas: Isn't this just `collect(eachcol(...))`?

ErikQQY: Yeah, it's just ```collect```, fixed

PR #152, opened 2023-12-19
Upper bound diffeqbase and patch bandind error

avik-pal:
We were relying on the `AbstractArray{<:AbstractArray}` for specifying initial conditions. This is the last version to support that. We will transition to a `VectorOfArray` from v6


avik-pal:
Don't merge I will look into the NLLS, it is most likely LM that is causing all the trouble. LM seems to be robust but also works worse than TR in practice

ChrisRackauckas (approved):

PR #159, opened 2024-01-10
Add COLNEW algorithm

ErikQQY:
Add COLNEW method from ODEInterface.jl.

1. The original COLNEW method supports multi-points BVP, its boundary condition must be defined as:
For example: if in time span $t = [-1, 1]$ we have boundary conditions $u(-1)=1, u'(0)=2.5, u(1)=5$, (here suppose `u[1]` is $u$, `u[2]` is $u'$)

```julia
zeta = [-1, 0, 1] # points in time span where we want to specify conditions
function bc(i, u, bc)
    if i == 1
        bc[1] = u[1] - 1.0
    elseif i == 2
        bc[1] = u[2] - 2.5
    elseif i == 3
        bc[1] = u[1] - 5.0
    end
end
......
sol, retcode, stats = colnew(tspan, orders, zeta, rhs, Drhs, bc, Dbc, nothing, opt)
```

the multi-points boundary condition definition in COLNEW makes it hard to convert to the API we are using here, so this PR only implemented the two-point case.

2. The initial guess is set to nothing because we support initial guess as initial guess value(set `u0` as `Vector`) during problem construction in BoundaryValueDiffEq.jl, whereas the initial guess in COLNEW can be [either `nothing` or the returned value from earlier COLNEW call](https://github.com/luchr/ODEInterface.jl/blob/master/doc/SolverOptions.md#guess)

### DIFF
--- a/ext/BoundaryValueDiffEqODEInterfaceExt.jl
+++ b/ext/BoundaryValueDiffEqODEInterfaceExt.jl
@@ -4,14 +4,17 @@ using SciMLBase, BoundaryValueDiffEq, ODEInterface
 import SciMLBase: __solve
 import ODEInterface: OptionsODE, OPT_ATOL, OPT_RTOL, OPT_METHODCHOICE, OPT_DIAGNOSTICOUTPUT,
     OPT_ERRORCONTROL, OPT_SINGULARTERM, OPT_MAXSTEPS, OPT_BVPCLASS, OPT_SOLMETHOD,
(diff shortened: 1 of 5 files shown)


ChrisRackauckas:
It would be nice to get some kind of side channel for its BC definition just to make benchmarking easier. I assume any translation like that would have to be done manually though, i.e. have inside of the `colnew` algorithm some `bc_func` option that overrides the BC with the manual translation

ErikQQY:
👍Making `bc_func` inside `colnew` to override BC with manual translation is a good idea.

PR #169, opened 2024-02-25
Update DEStats in ODEInterface extension

ErikQQY:
```DEStats``` updated in https://github.com/SciML/SciMLBase.jl/pull/630

This change is required for https://github.com/SciML/SciMLBenchmarks.jl/pull/880


ChrisRackauckas:
We probably shouldn't be using the inner constructor directly, that's the root cause of the issue here, but this is fine for now.

Issue #43, opened 2019-05-11
BigFloat support for TwoPointBVProblem using MIRK4


seadra:
Using BigFloat with a TwoPointBVProblem and MIRK4 solver currently fails with the following error:

> ArgumentError: matrix type SparseArrays.SparseMatrixCSC{BigFloat,Int64}not supported. Try lu(convert(SparseMatrixCSC{Float64/ComplexF64,Int}, A)) for sparse floating point LU using UMFPACK or lu(Array(A)) for generic dense LU.

ChrisRackauckas:
We probably just need to add a dependency on GenericSVD.jl

seadra:
I naively tried using GenericSVD in my code but it didn't work, so I'm guessing that means something else :)

avik-pal:
I thought `BigFloat` would just work, but seems like the problem is more fundamental

```julia
julia> similar(rand(BigFloat, 2, 3))
2×3 Matrix{BigFloat}:
 #undef  #undef  #undef
 #undef  #undef  #undef
```

This makes all of the internal NonlinearSolve machinery fail because it often performs concatenation and such on a `similar`ed array before assigning values to it.

ChrisRackauckas:
For safety, just zero?

avik-pal:
yes probably that's the way to go, it just needs to be added at quite a few places in NonlinearSolve

avik-pal:
NonlinearSolve now supports bigfloats natively. Need to add a test for this here.

PR #178, opened 2024-03-31
Add Lobatto IIIa-c and Radau IIa solvers

axla-io:
## Features

- 17 new Fully Implicit Runge-Kutta (FIRK) solvers: Lobatto IIIa-c 2 to 5 stage versions, and Radau IIa 1,2,3,5,7 - stage versions
- Choice of cache with nested nonlinear solve (better performance for large problems) or expanded cache (seems to be more robust)
- Adaptivity and interpolations
- Sparse Jacobians

## Issues with the code
Tests pass for the expanded cache, except for in some cases with the convergence tests for the higher order solvers, I think that the reason for this is that the error is low in absolute terms. I've done a check where the tests are marked as broken if the solution error is less than `1e-12`

Also there is a runtime dispatch error when calling the `JET` tests. I have commented the lines in `src/solve/firk.jl` where the error occurs. I couldn't figure out how to resolve them.

For the nested cache, in addition to the tests that the expanded one doesn't pass, it also seem to have poor performance on `NLLNS` and `Vector of vector initials` tests. I haven't found a solver for which the nested ones converges there. For now these tests are commented out.

Also, the nested cache doesn't support `ForwardDiff` yet.

## Checklist

- [x ] Appropriate tests were added
- [x ] Any code changes were done in a way that does not break public API
- [x ] All documentation related to code changes were updated
  [COLPRAC](https://github.com/SciML/COLPRAC). (_I commented out tests for nested cache as a WIP thing_)
- [ ] Any new documentation only uses public API (_I don't understand what this means :(_ )


src/adaptivity.jl
```diff
@@ -1,85 +1,207 @@
...
+	info = ReturnCode.Success
+
+	ŝ = [maximum(abs, d) for d in defect]  # Broadcasting breaks GPU Compilation
+	ŝ .= (ŝ ./ abstol) .^ (T(1) / (order + 1))
+	r₁ = maximum(ŝ)
+	r₂ = sum(ŝ)
+	r₃ = r₂ / (N - 1)
+
+	n_predict = round(Int, (safety_factor * r₂) + 1)
+	n = N - 1
```
avik-pal, line 171: I will review more carefully later but Tabs need to be changed to spaces

avik-pal (changes requested):

ChrisRackauckas:
> Tests pass for the expanded cache, except for in some cases with the convergence tests for the higher order solvers, I think that the reason for this is that the error is low in absolute terms. I've done a check where the tests are marked as broken if the solution error is less than 1e-12

Can you plot what those look like? for higher order you likely need to increase the dts

avik-pal:
Also let's merge the tests for MIRK and the other methods, except Shooting. The only reason Shooting is separate, is that the other methods currently don't support interpolation in the solution.

ErikQQY:
> Also let's merge the tests for MIRK and the other methods, except Shooting. The only reason Shooting is separate, is that the other methods currently don't support interpolation in the solution.

@avik-pal didn't the MIRK solvers already incorporate interpolations in https://github.com/SciML/BoundaryValueDiffEq.jl/pull/112?

avik-pal:
No I meant the interpolation in the residual function

ErikQQY:
> No I meant the interpolation in the residual function

Oh, indeed, there are no interpolations in the residual functions for MIRK methods right now.

src/adaptivity.jl
axla-io: Thanks! I changed this with the latest commit

axla-io:
@ChrisRackauckas This is how it looks like. I tried increasing / decreasing dts with LobattoIIIb stage 5, but to no avail.
lobattoIIIa_p3
lobattoIIIa_p7
lobattoIIIb_p3
lobattoIIIb_p7
lobattoIIIc_p3
lobattoIIIc_p7
radau_p3
radau_p7

ChrisRackauckas:
That looks like the opposite direction that the convergence plot recipe normally does. Is this not just `plot(sim)`?

axla-io:
@ChrisRackauckas Oops, I wasn't aware of that one. Here are the highest order solvers plotted with `plot(sim)`
lobattoIIIa_p3
lobattoIIIa_p7
lobattoIIIb_p3
lobattoIIIb_p7
lobattoIIIc_p3
lobattoIIIc_p7
radau_p3
radau_p7

ChrisRackauckas:
It's just below Float64 tolerance. Can you change it to BigFloat in state and time and calculate the convergence with arbitrary precision?

axla-io:
@ChrisRackauckas I get an `ERROR: UndefRefError: access to undefined reference` when using `BigFloat` for these solvers, and also `MIRK`. It seems to come from the autodiff and the error is there both for sparse and dense AD, both forward mode and finitediff

avik-pal:
BigFloat semantics are somewhat incompatible with how some of the solvers are written.

https://github.com/SciML/BoundaryValueDiffEq.jl/issues/43#issuecomment-2016859268

avik-pal:
> BigFloat semantics are somewhat incompatible with how some of the solvers are written.

This was fixed upstream; forgot to mention it here.

ErikQQY:
Current implementation passes basic tests for a BVP solver, e.g. discrete solving, adaptivity, interpolations, and plotting are all good now.

Current issues of this PR:
1. BigFloat semantics(I tried to solve the BigFloat issue for both MIRK and FIRK by following the patch in NonlinearSolve.jl, but seemed the current diffcache used in doesn't compatible with bigfloat, need to do some further searching)
2. For some large systems, such as the overconstrained and underconstrained tests in nonlinear least square tests fail

@ChrisRackauckas  @avik-pal, could you review this PR and see what else is still missing to get this PR merged? I will try to address these issues

avik-pal:
> BigFloat semantics(I tried to solve the BigFloat issue for both MIRK and FIRK by following the patch in NonlinearSolve.jl, but seemed the current diffcache used in doesn't compatible with bigfloat, need to do some further searching)

What is the error?

ChrisRackauckas:
How about switching to a lazy buffer cache to get this completed?

ErikQQY:
> How about switching to a lazy buffer cache to get this completed?

I have tried both `LazyBufferCache` and `FixedSizeDiffCache`, but none of them seems to support ForwardDiff AD when we are using BigFloat in BoundaryValueDiffEq.jl, for example the `LazyBufferCache` version https://github.com/SciML/PreallocationTools.jl/issues/72 still fail.

ErikQQY:
The current issue is that without using `DiffCache`, it is impossible to build such Dual number with BigFloat type to support ForwardDiff AD things. I don't think changing to `LazyBufferCache` or `GeneralLazyBufferCache` can solve this.

test/firk/expanded/nlls_tests.jl
```diff
@@ -0,0 +1,208 @@
+# Overconstrained and Underconstrained are taking too much time, skip them for now
+#=
```
ChrisRackauckas, line 2: should this just be a different test group?

ErikQQY: Yeah, but the CI takes a long time to test them so just commented them out for now.

avik-pal: An ugly way to do this is https://github.com/LuxDL/LuxLib.jl/blob/main/test/common_ops/dense_tests.jl. Create a module with the test and run it in separate groups, that way you get parallelism.

Also if the tests are too heavy we can move these to the buildkite cpus.

ErikQQY: Let's try putting them into a different test group

ErikQQY: The nlls tests for Lobatto and Radau methods take so much time😅: https://github.com/SciML/BoundaryValueDiffEq.jl/actions/runs/10447244427/job/28925868285?pr=178

Are there some issues with the methods or something else?

ChrisRackauckas: what's it spending all of that time doing?

avik-pal: Just to confirm this is not some issue with parallel testing, let's disable parallel testing for this PR

ErikQQY: Just figured out the problem, the last commit has changed the number of expanded solvers for nlls tests to 4*4=16, and the CI says nlls tests cost approximately 37minutes  `2270.1 secs (36.1% compile, 18.6% GC), 2.57 B allocs (314.128 GB)`😅. The previous nlls testing cases test all of the 68 FIRK methods, which would need a looooong time to finish😅.

ErikQQY:
The Lobatto and Radau solvers without nested nonlinear solving are working fine now, but those using nested nonlinear solving behave bad and don't converge on several test problems.

Maybe we can get those Lobatto and Radau without nested nonlinear solving in first?

ChrisRackauckas:
How come that doesn't show up as a test failure?

ErikQQY:
Those failed tests are commented out, the current CI only test FIRK solvers without nested nonlinear solving, but from local testing, the nested nonlinear solving choice only passes the basic test cases like affineness and linear problem tests.

ChrisRackauckas:
I believe @axla-io had said nested had issues earlier. Let's get this to merge where nested throws a warning where it's still experimental and its tests are properly set to `@test_broken`.

src/adaptivity.jl
```diff
@@ -12,6 +12,128 @@ After we construct an interpolant, we use interp_eval to evaluate it.
...
+
+    # Load interpolation residual
+    for jj in 1:stage
+        K[:, jj] = cache.y[ctr_y + jj].du
+    end
+
+    z₁, z₁′ = eval_q(yᵢ, 0.5, h, q_coeff, K) # Evaluate q(x) at midpoints
+    S_coeffs = get_S_coeffs(h, yᵢ, yᵢ₊₁, z₁, dyᵢ, dyᵢ₊₁, z₁′)
+
+    y .= S_interpolate(τ * h, S_coeffs)
```
ChrisRackauckas, line 61: Can this be done in-place to remove the allocation?

src/adaptivity.jl
```diff
@@ -12,6 +12,128 @@ After we construct an interpolant, we use interp_eval to evaluate it.
...
+    if lf > 1
+        h *= lf
+    end
+    τ = (t - mesh[j]) / h
+
+    (; f, M, p, ITU, TU) = cache
+    (; c, a, b) = TU
+    (; q_coeff, stage) = ITU
+
+    K = zeros(eltype(cache.y[1].du), M, stage)
```
ChrisRackauckas, line 34: unnecessary allocation?

src/adaptivity.jl
```diff
@@ -12,6 +12,128 @@ After we construct an interpolant, we use interp_eval to evaluate it.
...
+
+    nest_nlsolve_alg = __concrete_nonlinearsolve_algorithm(nest_prob, alg.nlsolve)
+    nestprob_p = zeros(T, cache.M + 2)
+
+    yᵢ = copy(cache.y[j].du)
+    yᵢ₊₁ = copy(cache.y[j + 1].du)
+
+    if iip
+        dyᵢ = copy(yᵢ)
+        dyᵢ₊₁ = copy(yᵢ₊₁)
```
ChrisRackauckas, line 88: There's a lot of allocations in these.

src/collocation.jl
```diff
@@ -61,3 +172,70 @@ end
...
+    end
+    return residuals
+end
+
+@views function Φ(fᵢ_cache, k_discrete, f!, TU::FIRKTableau{true}, y, u, p,
+        mesh, mesh_dt, stage::Int, cache)
+    (; c, a, b) = TU
+    (; nest_prob, alg, nest_tol) = cache
+
+    residuals = [similar(yᵢ) for yᵢ in y[1:(end - 1)]]
```
ChrisRackauckas, line 217: it defeats the purpose of in-place if you're creating new caches every step.

ChrisRackauckas:
Parts of the code need some work for performance. But let's not worry about that for now.

ChrisRackauckas:
Fix the deprecations and the `@test_broken`s and this is then good to merge.

ChrisRackauckas:
Hit a time limit? We really need to split this into different test groups.

ErikQQY:
Yeah, I will split them into different test groups

ChrisRackauckas:
Is the Retest parallelism actually helpful? BLAS will multithread, so we may be slamming the threads on any large enough problem. I don't think that part makes sense in the BVP library, or ML @avik-pal ?

avik-pal:
> Is the Retest parallelism actually helpful? BLAS will multithread, so we may be slamming the threads on any large enough problem. I don't think that part makes sense in the BVP library, or ML @avik-pal ?

Are any of the tests here heavy BLAS? In the ML context, none of the Lux tests are heavy enough to  cause issues with oversubscribing, and it helps bring down the build times to about 25 mins (compared to > 1hr) on decent machines (the cuda ones on buildkite). Not the most helpful for github actions.

That said for testing heavy workloads (both compute or memory), I do disable parallelism in testing. I still keep ReTestItems because it lets you isolate the testitem that fails and for testing Enzyme which can segfault/assert crash julia it will just crash the worker, so the test shows up as failed rather than the main process crashing.

ChrisRackauckas:
The LU/QR will all multithread above 200x200, which I presume we must be hitting?

avik-pal:
https://github.com/SciML/BoundaryValueDiffEq.jl/actions/runs/10617623160/job/29430931485?pr=178#step:6:397 the AD isn't robust to rank deficient matrices. This probably needs to be handled upstream

ErikQQY:
> https://github.com/SciML/BoundaryValueDiffEq.jl/actions/runs/10617623160/job/29430931485?pr=178#step:6:397 the AD isn't robust to rank deficient matrices. This probably needs to be handled upstream

From local  testing, RadauIIa5 just error with GaussNewton, the other nonlinear solvers are just fine, I have just skipped this combination for now

ChrisRackauckas:
That's the least stable one. Not surprised. It should really default to using a trust region method for this reason.

ChrisRackauckas:
Oh no I thought it was going to merge 😅

ErikQQY:
Yeah it was going to, but CI seemed to have a different opinion, let’s see how about now😅

ErikQQY:
Finally ready to go🎉🎉

ChrisRackauckas:
Amazing, and thanks @axla-io too.

PR #188, opened 2024-05-26
Migrating to new ADTypes

avik-pal:
Supercedes #187 #181


ChrisRackauckas (approved):

avik-pal:
Failures are harmless for end users, so I will merge this to unblock the upstream stuff. I will handle the failures in a followup PR

PR #189, opened 2024-06-07
COLNEW supports multi-points BVP

ErikQQY:
Now the COLNEW wrapper can handle multi-points BVP

### DIFF
--- a/test/shooting/basic_problems_tests.jl
+++ b/test/shooting/basic_problems_tests.jl
@@ -196,9 +196,9 @@ end
     flow_bvp = BVProblem{true}(flow_in_a_channel!, bc_flow!, u0, tspan, p)
 
     for solver in [
-        Shooting(AutoVern7(Rosenbrock23()),
+        Shooting(AutoTsit5(Rodas4P()),
             NewtonRaphson(; autodiff = AutoForwardDiff(; chunksize = 8))),
-        MultipleShooting(10, AutoVern7(Rosenbrock23()),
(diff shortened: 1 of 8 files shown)


ChrisRackauckas:
Need tests passing again to finish this.

test/shooting/basic_problems_tests.jl
```diff
@@ -35,12 +38,13 @@
         @test SciMLBase.successful_retcode(sol)
         @test norm(sol.resid, Inf) < 1e-8
 
+        JET_SKIP[i] && continue
         @test_opt target_modules=(BoundaryValueDiffEq,) solve(
             bvp1, solver; abstol = 1e-8, reltol = 1e-8,
-            odesolve_kwargs = (; abstol = 1e-6, reltol = 1e-3))
+            odesolve_kwargs = (; abstol = 1e-6, reltol = 1e-3)) broken=JET_BROKEN[i]
```
ChrisRackauckas, line 44: what's wrong with the JET tests?

ErikQQY: JET skipping here was introduced by @avik-pal, since most of the JET tests are marked as broken, I am thinking maybe we can get rid of them? If so, the only thing blocking this PR is #191

ErikQQY: @avik-pal What's left to deal with is the `MultipleShooting` for Ray tracing oop example error about `f` being non-automatic-differentiable(iip tests working fine), which is weird, I remember before the ADTypes upgrading they were working fine. Any idea how to fix this?

PR #194, opened 2024-07-04
[WIP] Add MIRKN methods

ErikQQY:
2nd, 4th and 6th order monotonic implicit runge-kutta-nystrom methods for second-order BVPs

It is still a WIP, I am working on setting up the jacobian and jac_prototype of the loss function, this should be similar to what we have done to MIRK methods, but the current implementation still has some problems, mainly about the banded matrix setup.


ErikQQY:
Too much conflicts here, will open a new one

PR #195, opened 2024-07-13
Fix CI

ErikQQY:
JET CI seems failing because the possible runtime dispatch in NonlinearSolve.jl, so this PR only tests if JET fails for BoundaryValueDiffEq.jl.

Screenshot from 2024-07-13 16-24-23

### DIFF
--- a/benchmark/simple_pendulum.jl
+++ b/benchmark/simple_pendulum.jl
@@ -18,18 +18,31 @@ function bc_pendulum!(residual, u, p, t)
     return nothing
 end
 
+function bc_pendulum_mirk!(residual, u, p, t)
+    residual[1] = u[end ÷ 2][1] + π / 2
+    residual[2] = u[end][1] - π / 2
+    return nothing
+end
+
 function simple_pendulum(u, p, t)
     g, L, θ, dθ = 9.81, 1.0, u[1], u[2]
     return [dθ, -(g / L) * sin(θ)]
 end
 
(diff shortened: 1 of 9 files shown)


ErikQQY:
CI tests for `Shooting` methods in `basic_problems_tests` are weird here, I got all the tests passing in my local environment🤔:

Screenshot from 2024-07-17 23-07-11

ChrisRackauckas:
What's that return code?

ErikQQY:
> What's that return code?

CI hit the Stalled return code in one of the Ray tracing tests, but the testing in local environment is fine.

ChrisRackauckas:
format and we can merge.

ChrisRackauckas:
the benchmark code looks like it needs an update too

ErikQQY:
The benchmark suite for MIRK methods failed because of the current indexing in boundary conditions for shooting methods and MIRK methods are different, e.g. MIRK doesn't support `res[1]=sol(t1)` but shooting methods does, which would require constructing solutions during boundary condition loss evalution like what we did in `Shooting`:

https://github.com/SciML/BoundaryValueDiffEq.jl/blob/20f27f44a43106175c439c2c4488395fc9a82d1b/src/solve/multiple_shooting.jl#L326-L336

Temporarily added different types of benchmarking problems, I will open another PR to unify the indexing in boundary conditions for MIRK and shooting methods.

PR #198, opened 2024-07-22
Fix benchmarks CI

ErikQQY:
Nonlinear solvers should be specified when using `Shooting` methods


ErikQQY:
This PR should be ready to go

avik-pal:
Seems to fail?

ErikQQY:
I think maybe the CI will work only after PR is merged? I tested locally and the benchmarking scripts are working:

image

ChrisRackauckas:
let's try it

PR #201, opened 2024-07-27
Fix StaticArrays support for MIRK methods

ErikQQY:
Fix: #193

### DIFF
--- a/test/mirk/mirk_basic_tests.jl
+++ b/test/mirk/mirk_basic_tests.jl
@@ -259,3 +259,29 @@ end
         (0, pi / 2), pi / 2; bcresid_prototype = (zeros(1), zeros(1)))
     @test SciMLBase.successful_retcode(solve(bvp5, MIRK4(), dt = 0.05).retcode)
 end
+
+@testitem "Compatibility with StaticArrays" begin
+    using StaticArrays
+    const g = 9.81
+    L = 1.0
+    tspan = (0.0, pi / 2)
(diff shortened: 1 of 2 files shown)


test/mirk/mirk_basic_tests.jl
```diff
@@ -259,3 +259,29 @@ end
...
+
+    function bc2a!(resid_a, u_a, p) # u_a is at the beginning of the time span
+        resid_a[1] = u_a[1] + pi / 2 # the solution at the beginning of the time span should be -pi/2
+    end
+    function bc2b!(resid_b, u_b, p) # u_b is at the ending of the time span
+        resid_b[1] = u_b[1] - pi / 2 # the solution at the end of the time span should be pi/2
+    end
+
+    bvp_SA = TwoPointBVProblem(
+        simplependulum!, (bc2a!, bc2b!), MVector{2}([pi / 2, pi / 2]),
```
ChrisRackauckas, line 283: Test SArray?

ErikQQY: Since the internal of MIRK solver involves a bunch of in-place array and cache mutating in both iip and oop problems, for example, `recursive_unflatten!`, `__append_similar` etc., SArray is not supported. To fix this, it would require a mutable cache and a bunch of rewrites of the current adaptivity routine, should we do the refactor to support SArray?

ChrisRackauckas:
I guess that's fine for now.

PR #204, opened 2024-07-31
ci: update tests workflow to use centralized reusable workflow and auto-suggest formatting changes on PRs

thazhemadam:
Follow up to #183.


ChrisRackauckas:
@ErikQQY is it sometimes an array sometimes an AbstractVectorOfArray? We should probably always make it an AbstractVectorofArray

ErikQQY:
> @ErikQQY is it sometimes an array sometimes an AbstractVectorOfArray? We should probably always make it an AbstractVectorofArray

We always use `Vector{Vector}}` type in the nonlinear problems constructing and solving, should we change all of them to the `AbstractVectorOfArray` type instead? That could make the recursive flatten and unflatten of a `Vector{Vector}` much easier.

ChrisRackauckas:
Yes, Make it an AbstractDiffEqArray, since there's time associated with it as well.

ChrisRackauckas:
@ErikQQY can you help with the completion of this?

ErikQQY:
Sure, I will help to get this complete

ErikQQY:
The errors in tthis PR are mostly caused by the deprecation of indexing methods in RecursiveArrayTools.jl:

There are two ways of specifying boundary conditions when constructing a `BVProblem`, e.g. `bc(u,p,t)=[u[1][1]]` or `bc(u,p,t)=[u(10.0)[1]]` before RecursiveArrayTools.jl deprecate `u[1]` for `u.u[1]` or `u[:,1]` kind indexing, both of them were working fine, but now the centralized CI errors are mostly caused by the deprecation of this API.

So here is the solution:

We need to change the convention of defining boundary conditions as `bc(u,p,t)=[u.u[1][1]]` or `bc(u,p,t)=[u[:,1][1]]` now for this PR to continue. **This may involve lots of tests, examples, and docs change.**

ChrisRackauckas:
> We need to change the convention of defining boundary conditions as bc(u,p,t)=[u.u[1][1]] or bc(u,p,t)=[u[:,1][1]] now for this PR to continue. This may involve lots of tests, examples, and docs change.

Yes, that's what needs to happen everywhere.

src/default_nlsolve.jl
```diff
@@ -18,7 +18,7 @@ function __FastShortcutBVPCompatibleNLLSPolyalg(
                 linsolve, precs, disable_geodesic = Val(true), autodiff, kwargs...),
             TrustRegion(; concrete_jac, linsolve, precs, autodiff, kwargs...),
             GaussNewton(; concrete_jac, linsolve, precs,
-                linesearch = LineSearchesJL(; method = BackTracking()),
+                linesearch = NonlinearSolve.LineSearchesJL(; method = BackTracking()),
```
ChrisRackauckas, line 21:
```suggestion
                linesearch = BackTracking(),
```

src/default_nlsolve.jl
```diff
@@ -33,7 +33,8 @@ function __FastShortcutBVPCompatibleNonlinearPolyalg(
     else
         algs = (NewtonRaphson(; concrete_jac, linsolve, precs, autodiff),
             NewtonRaphson(; concrete_jac, linsolve, precs,
-                linesearch = LineSearchesJL(; method = BackTracking()), autodiff),
+                linesearch = NonlinearSolve.LineSearchesJL(; method = BackTracking()),
```
ChrisRackauckas, line 36:
```suggestion
                linesearch = BackTracking(),
```

ChrisRackauckas:
@avik-pal do you know why this ambiguity is still reported? I don't see it actually being hit?

avik-pal:
https://github.com/SciML/BoundaryValueDiffEq.jl/blob/master/test/misc/qa_tests.jl#L5 needs to be removed

ErikQQY:
The Windows CI test looks weird, all other tests are working fine

PR #205, opened 2024-08-06
Use VectorOfArray for arrays in MIRK

ErikQQY:
Part of #204 

Use `VectorOfArray` to replace `Vector{Vector}`.

The reason why using `VectorOfArray` instead of `DiffEqArray` is that some arrays, like `y₀` need extension when we halve the mesh and start a new nonlinear solving, and such extension for `DiffEqArray` is troublesome.


ChrisRackauckas:
lots of test failures here.

PR #208, opened 2024-08-19
Boundary conditions should always use solution object

### LINKED ISSUE #107: README example should use interpolation instead of indexing
While trying ```MIRK``` solvers and ```Shooting``` solvers with example problem in README, the two kinds of solvers gave two different solutions:

```julia
using BoundaryValueDiffEq, OrdinaryDiffEq
function simplependulum!(du, u, p, t)
    θ = u[1]
    dθ = u[2]
    du[1] = dθ
    du[2] = -9.81 * sin(θ)
end
function bc!(residual, u, p, t)
    residual[1] = u[end ÷ 2][1] + pi / 2
    residual[2] = u[end][1] - pi / 2
end
tspan = (0.0, pi / 2)
prob = BVProblem(simplependulum!, bc!, [pi / 2, pi / 2], tspan)
mirk_sol = solve(prob, MIRK4(), dt = 0.05)
shooting_sol = solve(prob, Shooting(Tsit5()), dt=0.05)
```

Solution from MIRK(**Correct**):

mirk_sol

Solution from Shooting(**Wrong**):

shooting_sol

### LINKED ISSUE #185: Indexing in boundary conditions
What is the current best way of doing indexing in the boundary conditions?
In the following test, evaluating the bc with a solution gives a deprecation warning,
but if you follow deprecation advice, the same bc function can no longer be used for solving the problem,
since the solver seems to use an Array of Array for evaluation the bc, and not a DiffEqArray.

https://github.com/SciML/BoundaryValueDiffEq.jl/blob/master/test/mirk/nlls_tests.jl#L11-L25
```jl
  Test threw exception
  Expression: norm(bc1(sol, nothing, tspan), Inf) < 0.01
  `Base.getindex(A::AbstractDiffEqArray, i::Int)` is deprecated, use `Base.getindex(A, :, i)` instead.
  Stacktrace:
```

ErikQQY:
Fix #107 
Fix #185

We should always pass a solution type to boundary conditions, and with this PR, the definition of boundary conditions for MIRK methods and Shooting Methods is consistent. I used linear interpolation to interpolate the solution, not sure it is optimal, but it is working fine. Now there is only one way to define our boundary condition:

```julia
using BoundaryValueDiffEq
tspan = (0.0, pi / 2)
function simplependulum!(du, u, p, t)
    θ = u[1]
    dθ = u[2]
    du[1] = dθ
    du[2] = -9.81 * sin(θ)
end
function bc!(residual, u, p, t)
    residual[1] = u(pi/4)[1] + pi / 2
    residual[2] = u(pi/2)[1] - pi / 2
end
prob = BVProblem(simplependulum!, bc!, [pi / 2, pi / 2], tspan)
sol = solve(prob, MIRK4(), dt = 0.05)
```


src/solve/mirk.jl
```diff
@@ -258,7 +258,7 @@ end
 
 @views function __mirk_loss(u, p, y, pt::StandardBVProblem, bc::BC, mesh, cache) where {BC}
     y_ = recursive_unflatten!(y, u)
-    soly_ = VectorOfArray(y_)
+    soly_ = MIRKSol(y_, mesh, cache)
```
ChrisRackauckas, line 261: Just pass the ODESolution itself. You've already constructed it and are updated the values?

ChrisRackauckas:
>  used linear interpolation to interpolate the solution, not sure it is optimal, but it is working fine

It's not optimal. Each tableau has an associated interpolation we should add to the solution.

src/solve/mirk.jl
ErikQQY: Yeah, I just created a simple solution type here, since what we need here is only the interpolation of this solution, so it's probably not necessary to create a new `ODESolution` every time in loss function?

ErikQQY: Just using ODESolution here while we are evaluating boundary conditions

src/solve/mirk.jl
```diff
@@ -250,40 +250,40 @@ end
...
+    eval_bc_residual!((resida, residb), pt, bc!, soly_, p, mesh)
     recursive_flatten_twopoint!(resid, resids, cache.resid_size)
     return nothing
 end
 
 @views function __mirk_loss(u, p, y, pt::StandardBVProblem, bc::BC, mesh, cache) where {BC}
     y_ = recursive_unflatten!(y, u)
-    soly_ = MIRKSol(y_, mesh, cache)
-    resid_bc = eval_bc_residual(pt, bc, soly_, p, mesh)
+    soly_ = SciMLBase.build_solution(cache.prob, cache.alg, mesh, y_, interp = MIRKInterpolation(cache.mesh, y_, cache))
```
ChrisRackauckas, line 261: you should be able to build the solution type once and just keep updating its values?

ErikQQY: Yeah, that's better

ChrisRackauckas:
This was done in a different PR right?

ErikQQY:
The changing of boundary condition definitions from `res[1] = u[:, end] - 1.0` to `res[1] = u(pi/2) - 1.0` will be done in this PR, though there are still some issues with building the solution type in the begining, this will be done very soon.

ChrisRackauckas:
This was completed in another PR right?

ErikQQY:
> This was completed in another PR right?

Not yet, we changed to use RecursiveArrayTools.jl in the evaluation of boundary conditions for collocation methods, but the solution interpolating in boundary conditions is not done yet.

PR #211, opened 2024-09-12
Refactor FIRK solvers

ErikQQY:
- Add precompilation options for Lobatto and Radau solvers, but precompiling all of them seems to take a long time, so just set them as false for now
- Previous tolerance for nested nonlinear solving is bad, change it to nothing as default so that NonlinearSolve.jl can choose automatically
- Some variables and functions in some functions are not used, so cleaned up a little
- Some functions are not optimal, improve their performance
- Unify the interpolant evaluation for MIRK and FIRK methods


ChrisRackauckas:
You should probably just split it to a subpackage, i.e. lib/BoundaryDiffEqFIRK, and then you can set it to precompile, but it does not have to be in the standard BounaryDiffEq load. The plan should be to do this split with all of the solvers.

ErikQQY:
Yeah, splitting BoundaryValueDiffEq.jl into subpackages like what we've done in OrdinaryDiffEq.jl is on my TODO list, so let's leave the precompile stuff into splitting part and get the others in?

ChrisRackauckas (approved):

PR #214, opened 2024-09-24
Split FIRK methods into subpackage

ErikQQY:
Continue #213

PR #215, opened 2024-09-29
Split Shooting methods into subpackage

ErikQQY:
Continue #214


ErikQQY:
Ready to go now

ErikQQY:
There are still some questions though:
1. Are we going to leave the extension algorithms from ODEInterface.jl in BoundaryValueDiffEq.jl or make them into a subpackage too?
2. I noticed [NonlinearSolve.jl is putting SimpleNonlinearSolve.jl into a subpackage](https://github.com/SciML/NonlinearSolve.jl/issues/456#issuecomment-2294889189), should we also do the same?

ChrisRackauckas:
> I noticed https://github.com/SciML/NonlinearSolve.jl/issues/456#issuecomment-2294889189, should we also do the same?

No harm either way. I think the key is to make a core that is very lean.

> Are we going to leave the extension algorithms from ODEInterface.jl in BoundaryValueDiffEq.jl or make them into a subpackage too?

Having them as extensions is fine if the core is lean.

PR #216, opened 2024-09-30
Core subpackage for BoundaryValueDiffEq

ErikQQY:
Should finish the splitting stuff


ErikQQY:
This PR is ready now

ext/BoundaryValueDiffEqODEInterfaceExt.jl
```diff
@@ -1,8 +1,13 @@
...
-import BoundaryValueDiffEq: __extract_u0, __flatten_initial_guess, __extract_mesh,
-                            __initial_guess_length, __initial_guess, __has_initial_guess
+using SciMLBase, BoundaryValueDiffEq, ODEInterface, RecursiveArrayTools, ConcreteStructs,
+      Setfield
+using PreallocationTools
+import SciMLBase: AbstractDiffEqInterpolation, StandardBVProblem, __solve, _unwrap_val
+include("../lib/BoundaryValueDiffEqCore/src/types.jl")
+include("../lib/BoundaryValueDiffEqCore/src/utils.jl")
+include("../lib/BoundaryValueDiffEqCore/src/algorithms.jl")
+include("../lib/BoundaryValueDiffEqCore/src/BoundaryValueDiffEqCore.jl")
```
ChrisRackauckas, line 10: why is this all directly included?

src/BoundaryValueDiffEq.jl
```diff
@@ -23,17 +23,9 @@ import SciMLBase: AbstractDiffEqInterpolation, StandardBVProblem, __solve, _unwr
...
-include("sparse_jacobians.jl")
-include("default_nlsolve.jl")
-
-function __solve(prob::BVProblem, alg::BoundaryValueDiffEqAlgorithm, args...; kwargs...)
-    cache = init(prob, alg, args...; kwargs...)
-    return solve!(cache)
-end
+include("../lib/BoundaryValueDiffEqCore/src/BoundaryValueDiffEqCore.jl")
+using ..BoundaryValueDiffEqCore
+include("../lib/BoundaryValueDiffEqCore/src/algorithms.jl")
```
ChrisRackauckas, line 28: why is this needed?

src/algorithms.jl
```diff
@@ -124,6 +104,8 @@ function BVPSOL(; bvpclass = 2, sol_method = 0, odesolver = nothing)
     return BVPSOL(bvpclass, sol_method, odesolver)
 end
 
+SciMLBase.allows_arbitrary_number_types(alg::BVPSOL) = true
```
ChrisRackauckas, line 107: this is a wrapped method, it's false

ext/BoundaryValueDiffEqODEInterfaceExt.jl
ErikQQY: Some functionalities in wrapper solvers require BoundaryValueDiffEqCore, but cannot directly use it since the core is not registered or can be directly `import ..BoundaryValueDiffEqCore` in the extension, so just temporarily including the necessary functionalities

src/BoundaryValueDiffEq.jl
ErikQQY: It's the wrapper algorithms in ODEInterfacExt require the basic `BoundaryVallueDiffAlgorithm` types to subtype wrapper algorithms

src/algorithms.jl
ErikQQY: But without specifying this trait, ```BVPSOL``` would error about using non-standard number type, which I believe is the initial guess that caused this error

```julia
ERROR: Non-standard number type (i.e. not Float32, Float64,
ComplexF32, or ComplexF64) detected as the element type
for the initial condition or time span. These generic
number types are only compatible with the pure Julia
solvers which support generic programming, such as
OrdinaryDiffEq.jl. The chosen solver does not support
this functionality. Please double check that the initial
condition and time span types are correct, and check that
the chosen solver was correct.

Solver: BVPSOL(bvpclass = 2, sol_method = 0)
u0 type: Vector{Vector{Float64}}
Timespan type: Tuple{Float64, Float64}

Some of the types have been truncated in the stacktrace for improved reading. To emit complete information
in the stack trace, evaluate `TruncatedStacktraces.VERBOSE[] = true` and re-run the code.
```

ext/BoundaryValueDiffEqODEInterfaceExt.jl
ChrisRackauckas: okay as long as it's temporary

src/BoundaryValueDiffEq.jl
ChrisRackauckas: but this should be removed by importing from core?

ErikQQY: yeah, it can be removed

src/algorithms.jl
ErikQQY: OK, I will make the changes in DiffEqBase then

.github/workflows/Tests.yml
```diff
@@ -34,6 +34,7 @@ jobs:
           - "FIRK(EXPANDED)"
           - "FIRK(NESTED)"
           - "WRAPPERS"
+        julia-version: [1.10]
```
ErikQQY, line 37: Not sure if this is the correct way to fall back to Julia 1.10? Seems the CI is still testing on 1.11.

ChrisRackauckas, line 37:
```suggestion
        julia-version: ['1.10']
```

ErikQQY:
It seems CI is still using Julia1.11 to run the tests

.github/workflows/Tests.yml
```diff
@@ -34,6 +34,7 @@ jobs:
           - "FIRK(EXPANDED)"
           - "FIRK(NESTED)"
           - "WRAPPERS"
+        julia-version: ['1.10']
     uses: "SciML/.github/.github/workflows/tests.yml@v1"
     with:
       group: "${{ matrix.group }}"
```
avik-pal, line 40:
```suggestion
      group: "${{ matrix.group }}"
      julia-version: ${{ matrix.version }}
```

.github/workflows/Tests.yml
```diff
@@ -34,6 +34,7 @@ jobs:
           - "FIRK(EXPANDED)"
           - "FIRK(NESTED)"
           - "WRAPPERS"
+        julia-version: ['1.10']
```
avik-pal, line 37:
```suggestion
        version: ['1.10']
```

Issue #76, opened 2022-07-06
Feature request: BVP-DAE support


JulianUlrich:
It would be nice, if solving of boundary value problems for implicit or semi-explicit DAEs was supported. Such solvers exist in FORTRAN (COLDAE).

ChrisRackauckas:
Agreed. First, we should add mass matrices to the BVProblem specification. It needs a BVPFunction, like the ODEFunciton, etc. to add such Jacobian and other metafunctions. That would get `mass_matrix`, so then linearly implicit BVP DAEs can be specified. Then yes we need to add that to the MIRK method.

JulianUlrich:
Nice, i am looking forward to it. I am surprised this has not been implemented in any commercial tool...

PR #217, opened 2024-10-03
Add BVDAE solvers

ErikQQY:
The BVDAE solvers are almost done now, but need further optimization in the style and improving its performance. The BVDAE solvers are similar to MIRK and FIRK methods, and their numerical workflow can be roughly divided into several steps:

(1) Build collocation equations with RK tableau. With Runge-Kutta coefficients at collocation points, build the collocation equations at each subinterval. After the collocation equations are built, we need to use [AlmostBlockDiagonals.jl](https://github.com/ErikQQY/AlmostBlockDiagonals.jl)) to solve the built linear systems, which has an almost block diagonal form, we can just factorize(pLU) and substitute to find the solutions.

(2) Construct a nonlinear system. Construct a nonlinear system using collocation equations. In the nonlinear solving process, residual control is deployed to obtain the numerical solution. The sparsity of the nonlinear system is also configured according to what we did in MIRK and FIRK.

(3) Error estimation. While we use defect control for error control in MIRK and FIRK, BVDAE solvers utilize a common error control mechanism, which compares the two solutions obtained from different mesh(solutions from original mesh and halved mesh) and estimates the errors.

(4) Refine mesh according to the error norm and degree of equidistribution. With the error estimation we get, we compute the degree of equidistribution of the approximate integral of function and accordingly decide to refine the mesh or just halve the mesh.

(5) Proceed for next iteration or successful return.

There are still several issues that need to be fixed:

- [ ] When solving multi-points BVDAE, we need to specify these fixed points explicitly, I am thinking about adding a problem type like `BVDAEProblem` like what we did for `DAEProblem`
- [ ] There are not enough test cases for boundary value differential-algebraic equations, so need further investigation for more solid test cases
- [ ] Any suggestions would be welcomed

Also implemented an interface for COLDAE in ODEInterface.jl, since the ODEInterface.jl is not actively maintained, just put the implementation in fork: https://github.com/ErikQQY/ODEInterface.jl

Fix: #76


lib/BoundaryValueDiffEqAscher/src/algorithms.jl
```diff
@@ -0,0 +1,60 @@
...
+            author={Uri M. Ascher and J. Christiansen and Robert D. Russell},
+            journal={Mathematics of Computation},
+            year={1979},
+            volume={33},
+            pages={659-679},
+            url={https://api.semanticscholar.org/CorpusID:121729124}
+        }
+        ```
+        """
+        @kwdef struct $(alg){N, J <: BVPJacobianAlgorithm} <: AbstractAscher
```
ChrisRackauckas, line 48: Why make each one a different type?

lib/BoundaryValueDiffEqAscher/src/ascher_tableaus.jl
```diff
@@ -0,0 +1,185 @@
...
+        p = coef[1, i]
+        for j in 2:k
+            p = p * t[k + 1 - j] + coef[j, i]
+        end
+        dm[i] = p
+    end
+end
+
+function rkbas!(s, coef, k::Integer, rkb)
+    # for more guass collocation points per subinterval
```
ChrisRackauckas, line 172: This is exactly the radau iia tableau?

ErikQQY: The tableaus come from the `COLNEW` paper: [A New Basis Implementation for a Mixed Order Boundary value ODE solver](https://epubs.siam.org/doi/pdf/10.1137/0908047), which I don't think is the radau iia tableau

ErikQQY:
> Are there tests with nonlinearleastsquaresproblem? I didn't see any.

Haven't found an appropriate test example for its case, previous work on this PR only ensures the "standard" BVDAE part is working fine, but I suspect COLDAE support over-constraints or under-constraints BVDAE

lib/BoundaryValueDiffEqAscher/src/ascher_tableaus.jl
ChrisRackauckas: Okay I dug into https://www.researchgate.net/publication/2435083_Collocation_Software_for_Boundary_Value_Differential-Algebraic_Equations a bit and talked with @YingboMa. It's a Gauss-Legendre quadrature https://en.wikipedia.org/wiki/Gauss%E2%80%93Legendre_quadrature. Instad of calling this Ascher we should just call it a FIRK that is GaussLegendre. 

Also, I don't really see the reason to be splitting the FIRK methods by different stage numbers, making that a runtime value could mean that we could adapt the order on the fly. But I think we can fix that in the future.

ChrisRackauckas:
Trying to understand the general system here. I think it's fine for the Ascher (Gauss-Legendre) to be kept separate from the other FIRK because they have a lot going on with their adaptivity algorithm and such, that's fine. Maybe we keep the name Ascher because of that but in the docstring mention it's Gauss-Legendre but with Ascher's adaptivity algorithm.

But the bigger point, for BVDAEs, is there anything that limits them to the GL methods? What I ask is, any FIRK method should work for the DAE in mass matrix form right? It would be good to test that this is method independent. I'm trying to see where this mass matrix is being used and I only see https://github.com/SciML/BoundaryValueDiffEq.jl/pull/217/files#diff-f206170122c6dd64b293fa8ee49c4b402d313b6ad26dfccc7bb1ad11624f91f9R65 so I'm a bit confused.

ErikQQY:
> Maybe we keep the name Ascher because of that but in the docstring mention it's Gauss-Legendre but with Ascher's adaptivity algorithm.

Good points, will add them then.

> But the bigger point, for BVDAEs, is there anything that limits them to the GL methods? What I ask is, any FIRK method should work for the DAE in mass matrix form right? It would be good to test that this is method independent. I'm trying to see where this mass matrix is being used and I only see https://github.com/SciML/BoundaryValueDiffEq.jl/pull/217/files#diff-f206170122c6dd64b293fa8ee49c4b402d313b6ad26dfccc7bb1ad11624f91f9R65 so I'm a bit confused.

Yeah, I can test if the FIRK tableaus can plug into this implementation but I suspected there would be extra work comparing that we can just replace FIRK tableaus with MIRK to get FIRK collocation methods for common BVPs. The current implementation follows the pattern of COLDAE, in which IIUC only needs to specify the differential vars, so the mass matrix here is only used for stating which equations should be the differential or algebraic equations.

ErikQQY:
Just replaced the Gauss-Legendre collocation points with Radau collocation points for the solving of BVDAEs with the current implementations, it's working now and turns out it's method independent, the reference collocation points[^1] being used are:
Alt Text

I think maybe for now we can keep on polishing this PR and get the basic structure of the Gauss-Legendre BVDAE solver in?

[^1]: Nonlinear Programming: Concepts, Algorithms and Applications to Chemical Processes, Chapter 10, Page 292

ChrisRackauckas:
> Just replaced the Gauss-Legendre collocation points with Radau collocation points for the solving of BVDAEs with the current implementations

What do you mean?

ErikQQY:
The current implementation employs Gauss-Legendre collocation points; if we switch to Radau collocation points, e.g. using the RadauIIA5 tableau, it remains capable of solving BVDAEs.

lib/BoundaryValueDiffEqAscher/src/utils.jl
```diff
@@ -0,0 +1,225 @@
...
+    end
+    return prototype, size.(prototype)
+end
+function __get_bcresid_prototype(::StandardBVProblem, prob::BVProblem, u)
+    prototype = prob.f.bcresid_prototype !== nothing ? prob.f.bcresid_prototype : zero(u)
+    return prototype, size(prototype)
+end
+
+# Custom pivot LU factorization and substitution for simple usage and
+# customized array types
```
ChrisRackauckas, line 39: This doesn't seem so good, it's not blocking based on cache sizes? What's the reason for this?

lib/BoundaryValueDiffEqAscher/src/utils.jl
```diff
@@ -0,0 +1,225 @@
...
+@views function recursive_flatten!(y::Vector, x::Vector{Vector{Vector{T}}}) where {T}
+    i = 0
+    for xᵢ in x
+        for xᵢᵢ in xᵢ
+            copyto!(y[(i + 1):(i + length(xᵢᵢ))], xᵢᵢ)
+            i += length(xᵢᵢ)
+        end
+    end
+    return nothing
+end
```
ChrisRackauckas, line 129: Is this not already in RecursiveArrayTools.jl?

lib/BoundaryValueDiffEqAscher/src/utils.jl
```diff
@@ -0,0 +1,225 @@
...
+        end
+    end
+    for k in n:-1:1
+        b[k] = b[k] / a[k, k]
+        @views __muladd!(-b[k], a[1:(k - 1), k], b[1:(k - 1)])
+    end
+end
+
+@inline function __muladd!(a, x, y)
+    y .= muladd(a, x, y)
```
ChrisRackauckas, line 117:
```suggestion
    y .= muladd.(a, x, y)
```

It's going to allocate without the other `.`?

ChrisRackauckas:
Left a few comments, but this just about ready?

lib/BoundaryValueDiffEqAscher/src/utils.jl
ErikQQY: The blocking factorization is used in the whole global collocation matrix(describe the discretized BVDAE system), but the simple pLU here is actually used for parameter condensation of each interior collocation matrix block.

lib/BoundaryValueDiffEqAscher/src/utils.jl
ErikQQY: Do you mean the flatten and unflatten of a `vector{vector}` or `vector{vector{vector}}` supported in RecursiveArrayTools.jl?

lib/BoundaryValueDiffEqAscher/src/utils.jl
ErikQQY: Yeah, just fixed

ErikQQY: Yeah, just fixed

ErikQQY:
> Left a few comments, but this just about ready?

Yeah, the correctness of solving a BVADE is ready I can be sure, but there are some other problems such as interfaces, and efficiency stuff need to be fixed.

lib/BoundaryValueDiffEqAscher/src/utils.jl
ChrisRackauckas: `Array`. I think think it has an unflatten though, but it would be good to add that as a documented utility there rather than baking it in.

ChrisRackauckas:
I do think some of the interface stuff should be done, but I don't want to block this. Let's get it in if tests pass and open some issues about the refactors I mentioned.

ErikQQY:
Sure thing, I will open issues about the next steps for polishing BVDAE solvers

PR #227, opened 2024-10-26
Fix precompilation on 1.11

ErikQQY:
#207

This should also be able to fix CI issues on 1.11


Project.toml
```diff
@@ -55,13 +55,13 @@ DiffEqDevTools = "2.44"
 FastAlmostBandedMatrices = "0.1.1"
 FastClosures = "0.3"
 ForwardDiff = "0.10.36"
-JET = "0.8"
```
avik-pal, line 58: any idea why compathelper didn't pick it up?

ErikQQY: I think CompatHelper only helps with compats for dependencies right? Since we are using JET in testing, CompatHelper would not be able to pick them.

ChrisRackauckas: We might have to do someething with subdirs? @ArnoStrouwen has done this thing before 
https://github.com/SciML/OrdinaryDiffEq.jl/blob/master/.github/workflows/CompatHelper.yml#L26

ChrisRackauckas:
@ErikQQY you should have the ability to tag already. Have fun with these 😅

Project.toml
ArnoStrouwen: https://github.com/JuliaRegistries/CompatHelper.jl/issues/166

PR #230, opened 2024-10-30
Add MIRKN methods

ErikQQY:
Superdeded #194 

Add 4th and 6th order Mono-Implicit Runge-Kutta-Nystrom methods for second-order boundary value problems

Although in the MIRKN paper: "MONO-IMPLICIT RUNGE–KUTTA–NYSTR  ̈OM METHODS WITH APPLICATION TO BOUNDARY VALUE ORDINARY DIFFERENTIAL EQUATIONS":
> A modified version of the MIRKDC software package has been developed to test the new methods(MIRKN methods).

But the paper doesn't explain the defect sample points deployed in MIRKN methods nor further details about the defect control adaptivity. So the current implementation only focused on the discrete solving process.


lib/BoundaryValueDiffEqMIRKN/test/mirkn_basic_tests.jl
```diff
@@ -0,0 +1,116 @@
...
+        SecondOrderBVProblem(odef2, bc, u0, tspan),
+        TwoPointSecondOrderBVProblem(odef1, (bc_a!, bc_b!), u0, tspan),
+        TwoPointSecondOrderBVProblem(odef2, (bc_a, bc_b), u0, tspan)]
+    dts = 1 .// 2 .^ (3:-1:1)
+    @testset "MIRKN$order" for order in (4, 6)
+        sim = test_convergence(
+                    dts, prob, mirkn_solver; abstol = 1e-8, reltol = 1e-8)
+        @test sim.errors[:final]≈order atol=testTol
+    end
+end
```
ErikQQY, line 115: The convergence tests for MIRKN4 and MIRK6 pass locally. Since the `ODEFunction(f, analytic = analytic_solution)` has arguments number checkings for `f`, how can we construct such a function with an analytic solution for `SecondOrderBVProblem`? I am thinking maybe adding a new function type to allow this in SciMLBase.jl?

ErikQQY: Nothing to worry about then, just forgot that I already added `DynamicalBVPFunction` a few months ago😅

ChrisRackauckas (approved):

PR #231, opened 2024-10-31
Update Project.toml

ErikQQY:
Tagbot keeps notifying:

> Triggering TagBot for merged registry pull request
This extra notification is being sent because I expected a tag to exist by now, but it doesn't.
You may want to check your TagBot configuration to ensure that it's running, and if it is, check the logs to make sure that there are no errors.

I think releasing a new patch of the parent package would help resolve this problem.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "BoundaryValueDiffEq"
 uuid = "764a87c0-6b3e-53db-9096-fe964310641d"
-version = "5.11.1"
+version = "5.11.2"
 
 [deps]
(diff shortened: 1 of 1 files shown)


Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "BoundaryValueDiffEq"
 uuid = "764a87c0-6b3e-53db-9096-fe964310641d"
-version = "5.11.1"
+version = "5.11.2"
```
ChrisRackauckas, line 3:
```suggestion
version = "5.12.0"
```

PR #238, opened 2024-11-07
Use NonlinearSolveFirstOrder subpackages and use new CI

ErikQQY:
Fix: #242
Fix: #243 
Fix: #244 

This PR should be able to fix lots of things:
- BoundaryValueDiffEq now only uses first-order nonlinear solvers from NonlinearSolveFirstOrder, so the default polyalgs need updates, and LineSearches.jl is removed in favor of LineSearch.jl.
- BoundaryValueDiffEq is now configured to pick up the latest sub-packages from the lib directory directly
- Removed the `__unsafe_nonlinearfunction` and use `NonlinearFunction` everywhere
- Ascher methods should expand `valstr` for the next iteration error estimate when the mesh is doubled, and a new test case is added
- Reduced the precompile workload for FIRK methods

PR #240, opened 2024-11-09
hotfix __unsafe_nonlinearfunction

ChrisRackauckas:
However, this function is really bad. Can we please just remove it? It's not correct to `nothing` the other arguments: it's simply not a correct thing to do!!! This function should not exist and in its current form it will break every few months. It's a really bad idea.

### DIFF
--- a/lib/BoundaryValueDiffEqCore/Project.toml
+++ b/lib/BoundaryValueDiffEqCore/Project.toml
@@ -1,7 +1,7 @@
 name = "BoundaryValueDiffEqCore"
 uuid = "56b672f2-a5fe-4263-ab2d-da677488eb3a"
 authors = ["Qingyu Qu <erikqqy123@gmail.com>"]
-version = "1.0.1"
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
The tests should be deving BoundaryValueDiffEqCore.

AayushSabharwal:
Yeah I just landed here as well. It's not type piracy, but it is undocumented internals and is the same vein of error that made BVDiffEq fail earlier when we added `discretes` to `ODESolution`. I think I'll give this package a quick look to see if I can spot any more potential issues. Better get rid of them now than have it break every so often.

AayushSabharwal:
BVDiffEq CI also needs fixing since we can't detect breakages from SciMLBase with the way it currently is 😅

ErikQQY:
> BVDiffEq CI also needs fixing since we can't detect breakages from SciMLBase with the way it currently is 😅

Yeah, I am fixing this in #238

Issue #218, opened 2024-10-05
Start using DifferentiationInterface


avik-pal:
Opening an issue to track what all is needed. With `Constant` being a thing now, this should be relatively straight-forward. Also NonlinearSolve update is done and released so this should unify the ecosystem.

Another thing I want to do it figure out if we can use an existing ADType and ditch the `BVPJacobianAlgorithm` or is that needed? My guess would be the latter because we want fine-grained control over how to deal with each part of the jacobian.

ChrisRackauckas:
Resizing is the difficult part though. Every time the time steps adapt, the cache will need to be resized.

gdalle:
> Resizing is the difficult part though. Every time the time steps adapt, the cache will need to be resized.

To start experimenting with this, I introduced a secret function in https://github.com/gdalle/DifferentiationInterface.jl/pull/521 with the right semantics. Right now it re-creates a cache from scratch every time, but we can optimize it for some backends and operators

ErikQQY:
Working on this, have a working implementation, will PR soon

gdalle:
Did you encounter any major hurdles? Did it lead to at least some amount of code simplification?

ErikQQY:
This issue can be safely closed now, supports for other AD backends will be tracked in #288

PR #258, opened 2024-11-21
Use DifferentiationInterface for the jacobian

ErikQQY:
Fix: #218

MIRK, FIRK, MIRKN, and Ascher methods are using DifferentiationInterface.jl to support sparse Jacobian exploitation, shooting methods are working in progress.

Changing SparseDiffTools.jl to DifferentiationInterfce.jl doesn't cost much pain, but the performance drops comparing with the current master branch, I suspect this maybe caused by some inappropriate matrix coloring and sparsity detection usage of DifferentiationInterface.jl and SparseMatrixColorings.jl in this implementation, for example, the `jac_prototype` only need the basic structure of jacobian, but didn't find similar functionalities in DI.jl, the construction of `ColoringProblem` and `**Algorithm` in SparseMatrixColorings.jl introduce allocations, so cc Guillaume for some advice.


gdalle:
Hi @ErikQQY, thanks for pinging me!

> the performance drops comparing with the current master branch

Could you maybe give me a standalone benchmark, even if it requires checking out the package from this branch?

> for example, the jac_prototype only need the basic structure of jacobian, but didn't find similar functionalities in DI.jl

If you already know the Jacobian structure, you want to use [`KnownJacobianSparsityDetector`](https://sciml.github.io/ADTypes.jl/stable/#ADTypes.KnownJacobianSparsityDetector) from ADTypes.jl as the `sparsity_detector` in `ADTypes.AutoSparse`.

> the construction of ColoringProblem and **Algorithm in SparseMatrixColorings.jl introduce allocations

That is unavoidable, but luckily you should be able to do it outside of the hot loop since it is encapsulated in the [preparation step](https://juliadiff.org/DifferentiationInterface.jl/DifferentiationInterface/stable/explanation/operators/#Preparation). The mantra of DI is "prepare once, reuse plenty of times". I'll check out your code but I suspect you may not be taking advantage of preparation.

gdalle:
Review in progress, but I think you may have forgotten to specify a `sparsity_detector` in `AutoSparse`.
Possible choices include:

- `TracerSparsityDetector` from [SparseConnectivityTracer.jl](https://github.com/adrhill/SparseConnectivityTracer.jl)
- `SymbolicsSparsityDetector` from [Symbolics.jl](https://github.com/JuliaSymbolics/Symbolics.jl)
- `KnownJacobianSparsityDetector` from ADTypes.jl

lib/BoundaryValueDiffEqAscher/src/ascher.jl
```diff
@@ -315,31 +315,50 @@ function __construct_nlproblem(cache::AscherCache{iip, T}) where {iip, T}
...
-    sd = alg.jac_alg.diffmode isa AutoSparse ? SymbolicsSparsityDetection() :
-         NoSparsityDetection()
-    ad = alg.jac_alg.diffmode
-    lossₚ = (iip ? __Fix3 : Base.Fix2)(loss, cache.p)
-    jac_cache = __sparse_jacobian_cache(Val(iip), ad, sd, lossₚ, lz, lz)
-    jac_prototype = init_jacobian(jac_cache)
+    resid_prototype = zero(lz)
+    diffmode = if alg.jac_alg.diffmode isa AutoSparse
+        AutoSparse(alg.jac_alg.diffmode;
+            coloring_algorithm = GreedyColoringAlgorithm(LargestFirst()))
```
gdalle, line 323:
```suggestion
        AutoSparse(alg.jac_alg.diffmode;
            sparsity_detector = Symbolics.SymbolicsSparsityDetector(),
            coloring_algorithm = GreedyColoringAlgorithm(LargestFirst()))
```
or you can use `TracerSparsityDetector` from SparseConnectivityTracer.jl

lib/BoundaryValueDiffEqCore/src/utils.jl
```diff
@@ -370,3 +370,7 @@ end
 @inline (f::__Fix3{F})(a, b) where {F} = f.f(a, b, f.x)
 
 # convert every vector of vector to AbstractVectorOfArray, especially if them come from get_tmp of PreallocationTools.jl
+
+get_dense_ad(::Nothing) = nothing
+get_dense_ad(ad) = ad
+get_dense_ad(ad::AutoSparse) = ADTypes.dense_ad(ad)
```
gdalle, line 376: what is this for?

gdalle:
Also I'm not sure why you perform differentiation with `get_dense_ad(diffmode)` instead of just `diffmode`?

lib/BoundaryValueDiffEqAscher/src/ascher.jl
```diff
@@ -315,31 +315,50 @@ function __construct_nlproblem(cache::AscherCache{iip, T}) where {iip, T}
...
+    else
+        alg.jac_alg.diffmode
+    end
+
+    jac_cache = if iip
+        DI.prepare_jacobian(
+            loss, resid_prototype, get_dense_ad(diffmode), lz, Constant(cache.p))
```
gdalle, line 330:
```suggestion
            loss, resid_prototype, diffmode, lz, Constant(cache.p))
```

lib/BoundaryValueDiffEqAscher/src/ascher.jl
```diff
@@ -315,31 +315,50 @@ function __construct_nlproblem(cache::AscherCache{iip, T}) where {iip, T}
...
+    else
+        DI.prepare_jacobian(loss, get_dense_ad(diffmode), lz, Constant(cache.p))
```
gdalle, line 332:
```suggestion
        DI.prepare_jacobian(loss, diffmode, lz, Constant(cache.p))
```

lib/BoundaryValueDiffEqAscher/src/ascher.jl
```diff
@@ -315,31 +315,50 @@ function __construct_nlproblem(cache::AscherCache{iip, T}) where {iip, T}
...
+    end
+
+    jac_prototype = if iip
+        DI.jacobian(
+            loss, resid_prototype, jac_cache, get_dense_ad(diffmode), lz, Constant(cache.p))
```
gdalle, line 337:
```suggestion
            loss, resid_prototype, jac_cache, diffmode, lz, Constant(cache.p))
```

lib/BoundaryValueDiffEqAscher/src/ascher.jl
```diff
@@ -315,31 +315,50 @@ function __construct_nlproblem(cache::AscherCache{iip, T}) where {iip, T}
...
+    else
+        DI.jacobian(loss, get_dense_ad(diffmode), lz, Constant(cache.p))
```
gdalle, line 339:
```suggestion
        DI.jacobian(loss, diffmode, lz, Constant(cache.p))
```

lib/BoundaryValueDiffEqAscher/src/ascher.jl
```diff
@@ -315,31 +315,50 @@ function __construct_nlproblem(cache::AscherCache{iip, T}) where {iip, T}
...
+    end
+
     jac = if iip
-        @closure (J, u, p) -> __ascher_mpoint_jacobian!(J, u, ad, jac_cache, lossₚ, lz)
+        @closure (J, u, p) -> __ascher_mpoint_jacobian!(
+            J, u, get_dense_ad(diffmode), jac_cache, loss, lz, cache.p)
     else
-        @closure (u, p) -> __ascher_mpoint_jacobian(jac_prototype, u, ad, jac_cache, lossₚ)
+        @closure (u, p) -> __ascher_mpoint_jacobian(
+            jac_prototype, u, get_dense_ad(diffmode), jac_cache, loss, cache.p)
```
gdalle, line 347:
```suggestion
            jac_prototype, u, diffmode, jac_cache, loss, cache.p)
```

lib/BoundaryValueDiffEqAscher/src/ascher.jl
```diff
@@ -315,31 +315,50 @@ function __construct_nlproblem(cache::AscherCache{iip, T}) where {iip, T}
...
+        DI.jacobian(
+            loss, resid_prototype, jac_cache, get_dense_ad(diffmode), lz, Constant(cache.p))
+    else
+        DI.jacobian(loss, get_dense_ad(diffmode), lz, Constant(cache.p))
+    end
+
     jac = if iip
-        @closure (J, u, p) -> __ascher_mpoint_jacobian!(J, u, ad, jac_cache, lossₚ, lz)
+        @closure (J, u, p) -> __ascher_mpoint_jacobian!(
+            J, u, get_dense_ad(diffmode), jac_cache, loss, lz, cache.p)
```
gdalle, line 344:
```suggestion
            J, u, diffmode, jac_cache, loss, lz, cache.p)
```

lib/BoundaryValueDiffEqAscher/src/ascher.jl
gdalle: in any case, without a sparsity detector, you will perform dense AD

lib/BoundaryValueDiffEqAscher/src/ascher.jl
gdalle: if you go back to the dense backend you won't see the benefits of `AutoSparse`

lib/BoundaryValueDiffEqCore/src/BoundaryValueDiffEqCore.jl
```diff
@@ -18,9 +18,10 @@ import Logging
 using NonlinearSolveFirstOrder: NonlinearSolvePolyAlgorithm
 import LineSearch: BackTracking
 import RecursiveArrayTools: VectorOfArray, DiffEqArray
-import SciMLBase: AbstractDiffEqInterpolation, StandardBVProblem, __solve, _unwrap_val
+import SciMLBase: AbstractBVProblem, AbstractDiffEqInterpolation, StandardBVProblem,
+                  __solve, _unwrap_val
 
-@reexport using ADTypes, NonlinearSolveFirstOrder, SparseDiffTools, SciMLBase
+@reexport using NonlinearSolveFirstOrder, SparseDiffTools, SciMLBase
```
gdalle, line 24: do you still need SparseDiffTools?

lib/BoundaryValueDiffEqFIRK/src/firk.jl
```diff
@@ -441,43 +441,67 @@ function __construct_nlproblem(
     L = length(resid_bc)
     resid_collocation = __similar(y, cache.M * (N - 1) * (stage + 1))
 
-    loss_bcₚ = (iip ? __Fix3 : Base.Fix2)(loss_bc, cache.p)
-    loss_collocationₚ = (iip ? __Fix3 : Base.Fix2)(loss_collocation, cache.p)
+    bc_diffmode = if jac_alg.bc_diffmode isa AutoSparse
+        AutoSparse(get_dense_ad(jac_alg.bc_diffmode);
+            sparsity_detector = ADTypes.TracerSparsityDetector(),
```
gdalle, line 446:
```suggestion
            sparsity_detector = SparseConnectivityTracer.TracerSparsityDetector(),
```

lib/BoundaryValueDiffEqFIRK/src/firk.jl
```diff
@@ -441,43 +441,67 @@ function __construct_nlproblem(
...
             # For underdetermined problems we use sparse since we don't have banded qr
-            colored_matrix = __generate_sparse_jacobian_prototype(
-                cache, cache.problem_type, y, y, cache.M, N)
             J_full_band = nothing
-            __sparsity_detection_alg(ColoredMatrix(
-                sparse(colored_matrix.M), colored_matrix.row_colorvec,
-                colored_matrix.col_colorvec))
+            colored_matrix = __generate_sparse_jacobian_prototype(
+                cache, cache.problem_type, y, y, cache.M,
+                N, get_dense_ad(jac_alg.nonbc_diffmode))
```
gdalle, line 464: what do you use this for?

lib/BoundaryValueDiffEqCore/src/utils.jl
gdalle: you seem to use it everywhere but if you go back from `AutoSparse(backend, ...)` to `backend` you lose the benefits of sparse AD

gdalle (changes requested):
I only reviewed the first few files because I think there are a few misunderstandings to clarify before going further

ChrisRackauckas:
There is a banded matrix with a known bandsize, and a potentially sparse set of rows on top. The current algorithm gets the color vector by knowing the bandsize analytical solution and does only sparsity on the extra rows. Then it does batched forward mode for the banded matrix and choice for the top (forward, reverse, etc). 

I think the difficulty here is that the representation of the coloring in DI makes it hard to do this by hand. The question is then how to get this mixed mode automatically or semi-automatically.

lib/BoundaryValueDiffEqCore/src/utils.jl
ErikQQY: Yeah, should use the dense ad from diffmode, just fixed

lib/BoundaryValueDiffEqCore/src/BoundaryValueDiffEqCore.jl
ErikQQY: The Shooting methods is still working on progress, which require SparseDiffTools.jl

adrhill:
Let me know if there are any bottlenecks in SCT's performance. We can most likely add additional methods for speed-ups.

ErikQQY:
> and you're interested in the Jacobian of f(x) = vcat(f1(x), f2(x))?

Yes

> Note that this would also be a good opportunity to try out AutoSparse(DI.MixedMode(...)) (but at the moment it works best with a random coloring order, because we're missing a way to https://github.com/gdalle/SparseMatrixColorings.jl/issues/122). Still rather experimental, and [the docs](https://juliadiff.org/DifferentiationInterface.jl/DifferentiationInterface/stable/api/#DifferentiationInterface.MixedMode) are pretty barebones.

Now we are using `BVPJacobianAlgorithm` with `bc_diffmode` and `nonbc_diffmode` to compute the Jacobian separately on the bc part(dense) and collocation part(sparse) in MRIK and FIRK algorithms, for these algorithms don't specify this we can just use `diffmode` for the Jacobian, I can try out `DI.MixedMode` for the almost banded situation

gdalle:
I don't understand, is the banded part dense or sparse?

ErikQQY:
Sorry, the banded part is dense, the bc part is sparse

gdalle:
What kind of storage do you use for this stacked Jacobian? Do you store both parts separately? Or do you put them both inside a `Matrix`? Or a `SparseMatrixCSC`?

ErikQQY:
Now we just store both parts in one matrix(it only matters in the multi-points case, since in two-points case the Jacobian is just a banded matrix which doesn’t need this kind of storage), for example the first several rows of our Jacobian are the sparse Jacobian of bc part(sparse because the evaluation of boundary conditions is only on a few points), and the other rows of the Jacobian have a banded structure. We store this special structure as almost banded matrix using [FastAlmostBandedMatrices.jl](https://github.com/SciML/FastAlmostBandedMatrices.jl).

gdalle:
Okay, we'll see how fast decompression is, but there might be some tweaks necessary there

gdalle:
How are we doing around here @ErikQQY? Need any more guidance?

ErikQQY:
@gallery Thanks a lot! Now I first need to ensure the correct syntax for the sparse Jacobian and the refactored solvers don’t break CI. Still have some subpackages to do, but should be ready very soon. After that, some reviews on the improvements and would be nice😄.

gdalle:
I wonder why this PR ends up adding more lines than it removes? Did we fail with our grand design of DI making life easier for package developers?

ErikQQY:
> I wonder why this PR ends up adding more lines than it removes? Did we fail with our grand design of DI making life easier for package developers?

This PR is kind of still working in progress, I haven't removed the sparse Jacobian stuff from SparseDiffTool.jl since the shooting methods exploit this in a special way(but will delete them once correctly configured!), and this is also mixed with some refactors so the git diff may not reflect the real status.

gdalle:
Okay thanks! But do keep it in mind while you make the changes: if DI is not making your life easier, there is probably something wrong that we can fix 😊

gdalle:
The failure for Ascher seems to be caused by the cache object not adapting to the element type of the input. During sparsity detection, SparseConnectivityTracer.jl passes a custom number type (not unlike ForwardDiff.jl), and there is apparently a forced conversion to `Float64` in your code somewhere?

ErikQQY:
> The failure for Ascher seems to be caused by the cache object not adapting to the element type of the input

Yeah, I am not quite satisfied with the current implementation of the collocation part of Ascher methods, and I am doing the overhaul these days. But I believe other methods like MIRK, FIRK, MIRKN, and Shooting methods are surely done now

gdalle:
Any feedback on things we could do in DI to make your life easier?

ErikQQY:
> Any feedback on things we could do in DI to make your life easier?

It is quite delightful to refactor from the SparseDiffTools.jl to DI, and the refactor does make the code simpler, but I guess DI is still a young package compared to others and I stumbled over several bugs(speaking of which, a new issue coming😅). But on the bright side, encountering problems and reporting them are also a way to help DI grow better🤠.

ErikQQY:
@gdalle @adrhill OK, the MIRK, FIRK, Shooting, and MIRKN methods are done now, the other methods need further optimization(need a redesign and don't involve DI, leave it out for now), so could you please give some advice on the current refactorization of changing SparseDiffTools.jl to DI? You can only focus on the changes in files: mirk.jl, firk.jl, mirkn.jl, single_shooting.jl, multiple_shooting.jl and their corresponding sparse_jacobians.jl where the matrix coloring structure is predetermined. Thanks in advance!!!

adrhill:
I have some big picture feedback that is might be more relevant to the design of ADTypes.jl:

It looks like you are internally overwriting the `sparsity_detector` in provided `AutoSparse` backends, e.g. in `mirk.jl` L315-347. Maybe ADTypes should provide a detector/trait that can be passed to packages like BoundaryValueDiffEq.jl such that they can provide their own patterns? Basically an empty `ADTypes.KnownJacobianSparsityDetector`.

lib/BoundaryValueDiffEqFIRK/src/firk.jl
```diff
@@ -441,43 +441,65 @@ function __construct_nlproblem(
...
-            __sparsity_detection_alg(__generate_sparse_jacobian_prototype(
-                cache, cache.problem_type, y, y, cache.M, N))
+            colored_result = __generate_sparse_jacobian_prototype(
+                cache, cache.problem_type, y, y, cache.M, N, jac_alg.nonbc_diffmode)
         end
+        AutoSparse(get_dense_ad(jac_alg.nonbc_diffmode);
+            sparsity_detector = ADTypes.KnownJacobianSparsityDetector(colored_result.A),
+            coloring_algorithm = ConstantColoringAlgorithm{ifelse(
+                ADTypes.mode(jac_alg.nonbc_diffmode) isa ADTypes.ReverseMode,
+                :row, :column)}(colored_result.A,
```
gdalle, line 477:
```suggestion
                :row, :column)}(sparsity_pattern(colored_result),
```
Fields of the coloring result are private, you should use the accessors defined by SparseMatrixColorings.jl (here and everywhere else)

lib/BoundaryValueDiffEqFIRK/src/firk.jl
```diff
@@ -441,43 +441,65 @@ function __construct_nlproblem(
...
+    else
+        DI.prepare_jacobian(loss_collocation, nonbc_diffmode, y, Constant(cache.p))
+    end
 
-    J_bc = zero(init_jacobian(cache_bc))
-    J_c = zero(init_jacobian(cache_collocation))
+    J_bc = if iip
+        DI.jacobian(loss_bc, resid_bc, cache_bc, bc_diffmode, y, Constant(cache.p))
+    else
+        DI.jacobian(loss_bc, bc_diffmode, y, Constant(cache.p))
```
gdalle, line 495:
```suggestion
        DI.jacobian(loss_bc, cache_bc, bc_diffmode, y, Constant(cache.p))
```
You can (and should) use the preparation result even when you compute the Jacobian out-of-place

lib/BoundaryValueDiffEqFIRK/src/firk.jl
```diff
@@ -511,40 +531,52 @@ function __construct_nlproblem(
...
             @view(cache.bcresid_prototype[(prod(cache.resid_size[1]) + 1):end]),
-            cache.M, N))
+            cache.M, N, jac_alg.diffmode)
+        AutoSparse(get_dense_ad(jac_alg.diffmode);
+            sparsity_detector = ADTypes.KnownJacobianSparsityDetector(colored_result.A),
+            coloring_algorithm = ConstantColoringAlgorithm{ifelse(
+                ADTypes.mode(jac_alg.bc_diffmode) isa ADTypes.ReverseMode, :row, :column)}(
+                colored_result.A,
+                ifelse(ADTypes.mode(jac_alg.nonbc_diffmode) isa ADTypes.ReverseMode,
+                    row_colors, column_colors)(colored_result)))
```
gdalle, line 557: This logic is a bit hard to parse, defining intermediate variables would help

lib/BoundaryValueDiffEqFIRK/src/firk.jl
```diff
@@ -730,51 +800,52 @@ end
...
         J, _, x, bc_diffmode, nonbc_diffmode, bc_diffcache, nonbc_diffcache, loss_bc::BC,
-        loss_collocation::C, resid_bc, resid_collocation, L::Int) where {BC, C}
-    sparse_jacobian!(@view(J[1:L, :]), bc_diffmode, bc_diffcache, loss_bc, resid_bc, x)
-    sparse_jacobian!(@view(J[(L + 1):end, :]), nonbc_diffmode,
-        nonbc_diffcache, loss_collocation, resid_collocation, x)
+        loss_collocation::C, resid_bc, resid_collocation, L::Int, p) where {BC, C}
+    DI.jacobian!(
+        loss_bc, resid_bc, @view(J[1:L, :]), bc_diffcache, bc_diffmode, x, Constant(p))
+    DI.jacobian!(loss_collocation, resid_collocation, @view(J[(L + 1):end, :]),
+        nonbc_diffcache, nonbc_diffmode, x, Constant(p))
```
gdalle, line 807: Note that if you ever work with `SparseMatrixCSC` formats (I think it is not the case here), views of sparse matrices are very inefficient

lib/BoundaryValueDiffEqFIRK/src/sparse_jacobians.jl
```diff
@@ -12,27 +12,34 @@ If the problem is a TwoPointBVProblem, then this is the complete Jacobian, else
...
 function __generate_sparse_jacobian_prototype(
-        ::FIRKCacheNested, ::StandardBVProblem, ya, yb, M, N)
+        ::FIRKCacheNested, ::StandardBVProblem, ya, yb, M, N, ad)
     fast_scalar_indexing(ya) ||
         error("Sparse Jacobians are only supported for Fast Scalar Index-able Arrays")
     J_c = BandedMatrix(Ones{eltype(ya)}(M * (N - 1), M * N), (1, 2M - 1))
-    return ColoredMatrix(J_c, matrix_colors(J_c'), matrix_colors(J_c))
+    problem = ColoringProblem(;
+        partition = ifelse((ADTypes.mode(ad) isa ADTypes.ReverseMode), :row, :column))
+    algo = GreedyColoringAlgorithm()
```
gdalle, line 21: At some spots in the code you use `LargestFirst()` while at others you do not, is it on purpose?

lib/BoundaryValueDiffEqFIRK/src/firk.jl
```diff
@@ -441,43 +441,65 @@ function __construct_nlproblem(
...
                 (block_size, block_size))
-            __sparsity_detection_alg(__generate_sparse_jacobian_prototype(
-                cache, cache.problem_type, y, y, cache.M, N))
+            colored_result = __generate_sparse_jacobian_prototype(
+                cache, cache.problem_type, y, y, cache.M, N, jac_alg.nonbc_diffmode)
         end
+        AutoSparse(get_dense_ad(jac_alg.nonbc_diffmode);
+            sparsity_detector = ADTypes.KnownJacobianSparsityDetector(colored_result.A),
+            coloring_algorithm = ConstantColoringAlgorithm{ifelse(
+                ADTypes.mode(jac_alg.nonbc_diffmode) isa ADTypes.ReverseMode,
```
gdalle, line 476: Note that there are other modes than `ForwardMode` and `ReverseMode`: ADTypes.jl also defines `SymbolicMode` and `ForwardOrReverseMode`. I don't know whether that is relevant here.

lib/BoundaryValueDiffEqMIRK/src/sparse_jacobians.jl
```diff
@@ -10,26 +10,33 @@ coloring.
...
     J₁ = length(ya) + length(yb) + M * (N - 1)
     J₂ = M * N
     J = BandedMatrix(Ones{eltype(ya)}(J₁, J₂), (M + 1, M + 1))
+    problem = ColoringProblem(;
+        partition = ifelse((ADTypes.mode(ad) isa ADTypes.ReverseMode), :row, :column))
+    algo = GreedyColoringAlgorithm()
     # for underdetermined systems we don't have banded qr implemented. use sparse
-    J₁ < J₂ && return ColoredMatrix(sparse(J), matrix_colors(J'), matrix_colors(J))
-    return ColoredMatrix(J, matrix_colors(J'), matrix_colors(J))
+    J₁ < J₂ && return coloring(sparse(J), problem, algo)
```
gdalle, line 40: The function `coloring` converts the matrix to `SparseMatrixCSC` anyway, so the `sparse` here is superfluous unless you need to recover a `SparseMatrixCSC` when you do `sparsity_pattern(coloring_result)` (aka `coloring_result.A`)

lib/BoundaryValueDiffEqMIRKN/src/mirkn.jl
```diff
@@ -96,71 +96,84 @@ function SciMLBase.solve!(cache::MIRKNCache{iip, T}) where {iip, T}
...
-    lossₚ = (iip ? __Fix3 : Base.Fix2)(loss, cache.p)
-    sd = alg.jac_alg.diffmode isa AutoSparse ? SymbolicsSparsityDetection() :
-         NoSparsityDetection()
-    ad = alg.jac_alg.diffmode
-    lz = reduce(vcat, cache.y₀)
-    jac_cache = __sparse_jacobian_cache(Val(iip), ad, sd, lossₚ, lz, lz)
-    jac_prototype = init_jacobian(jac_cache)
+    diffmode = if alg.jac_alg.diffmode isa AutoSparse
+        AutoSparse(get_dense_ad(alg.jac_alg.diffmode);
+            # Local because iszero require primal values
```
gdalle, line 118: This means that the sparsity pattern will not necessarily generalize across input points. To increase robustness, you probably want to evaluate it at `x = rand(n)` rather than, say, `x = zeros(n)`.

lib/BoundaryValueDiffEqShooting/src/multiple_shooting.jl
```diff
@@ -209,15 +233,25 @@ function __multiple_shooting_init_odecache(
...
 
 function __multiple_shooting_init_jacobian_odecache(
-        ensemblealg, prob, jac_cache, ::DiffCacheNeeded, alg, nshoots, u; kwargs...)
-    cache = jac_cache.cache
-    if cache isa ForwardDiff.JacobianConfig
-        xduals = reshape(cache.duals[2][1:length(u)], size(u))
+        ensemblealg, prob, jac_cache, ::DiffCacheNeeded,
+        diffmode, alg, nshoots, u; kwargs...)
+    if diffmode isa AutoSparse
+        xduals = reshape(jac_cache.pushforward_prep.xdual_tmp[1:length(u)], size(u))
```
gdalle, line 252: This is private API of DifferentiationInterface, it can change at any time without notice, so you shouldn't touch it. The details of preparation objects are not public-facing.

Why do you need to manipulate dual numbers directly?

lib/BoundaryValueDiffEqShooting/src/multiple_shooting.jl
```diff
@@ -209,15 +233,25 @@ function __multiple_shooting_init_odecache(
...
+    elseif diffmode isa AutoForwardDiff
+        xduals = reshape(jac_cache.config.duals[2][1:length(u)], size(u))
```
gdalle, line 254: Same here except that it is private API of ForwardDiff (less likely to break though)

gdalle (changes requested):

gdalle:
> Maybe ADTypes should provide a detector/trait that can be passed to packages like BoundaryValueDiffEq.jl such that they can provide their own patterns? Basically an empty ADTypes.KnownJacobianSparsityDetector.

What do you mean by that? Packages are already free to subtype `AbstractSparsityDetector` if they want

adrhill:
> > Maybe ADTypes should provide a detector/trait that can be passed to packages like BoundaryValueDiffEq.jl such that they can provide their own patterns? Basically an empty ADTypes.KnownJacobianSparsityDetector.
> 
> What do you mean by that? 

Maybe users should be provided the choice between:
* `UserKnownJacobianSparsityDetector`[^1]: The user defines the pattern. This essentially is the current `KnownJacobianSparsityDetector`, a struct that holds the known pattern.
* `PackageKnownJacobianSparsityDetector`[^1]: The user wants to let the package define the pattern. This would be an empty struct to dispatch on. This is what this PR does for any `AutoSparse` backend.

> Packages are already free to subtype `AbstractSparsityDetector` if they want

Of course, every package could their own version of `PackageKnownJacobianSparsityDetector`, just like they could all define their own `AutoEnzyme` variant, but providing a common interface for (sparse) AD it is the entire purpose of ADTypes.jl.

[^1]: Disregard the awful naming of the two types, I just wanted to make clear that the current `KnownJacobianSparsityDetector` is ambiguous as to whether (a) the user or (b) the package knows the pattern.

lib/BoundaryValueDiffEqFIRK/src/firk.jl
ErikQQY: Fixed

lib/BoundaryValueDiffEqFIRK/src/firk.jl
ErikQQY: Fixed. Just out of curiosity, what's the difference between using or not using the preparation result when calling `DI.jacobian` on the OOP function? Using the preparation can make Jacobian computing much faster?

lib/BoundaryValueDiffEqFIRK/src/sparse_jacobians.jl
ErikQQY: Fixed, all `GreedyColoringAlgorithm` are using the default

lib/BoundaryValueDiffEqMIRK/src/sparse_jacobians.jl
ErikQQY: Actually, we do need to recover a `SparseMatrixCSC` for the `KnownJacobianSparsityDetector`, so I think the `sparse` here is fine

lib/BoundaryValueDiffEqShooting/src/multiple_shooting.jl
ErikQQY: In Shooting methods, the loss function is the evaluation of an ODE integrator in boundary conditions, for example, after initializing an ODE problem, we would have an ODE integrator, and we use Newton methods to find the point where the residual satisfies our tolerance, in this process, we have to re-initialize the integrator(re-initializing points are iteration points) until Newton methods achieve convergence. OK, this procedure is basically the shooting methods, but we want to use AD to utilize the Jacobian, so the answer to the Jacobian of this kind of loss function is to build another integrator whose `u` is Dual number, and in the process of re-initialization, we need Dual number iteration points. Here is an example illustrating how shooting methods work:

```julia
using OrdinaryDiffEq, SciMLBase, ForwardDiff, DifferentiationInterface, ADTypes
const DI = DifferentiationInterface

backend = AutoForwardDiff()
function lorenz!(du, u, p, t)
    du[1] = 10.0(u[2] - u[1])
    du[2] = u[1] * (28.0 - u[3]) - u[2]
    du[3] = u[1] * u[2] - (8 / 3) * u[3]
end
u1 = [1.0; 0.0; 0.0]
tspan = (0.0, 100.0)
prob = ODEProblem(lorenz!, u1, tspan)
ode_cache = SciMLBase.__init(prob, Tsit5())
function loss(du, u, p)
    SciMLBase.reinit!(ode_cache, u)
    sol = solve!(ode_cache)
    du .= sol[end]
end
u1 = [2.0; 2.0; 3.0]
jac_cache = DI.prepare_jacobian(nothing, similar(u1), backend, u1)#ForwardDiff.JacobianConfig(nothing, u1, Chunk(2))
xduals = jac_cache.config.duals[2]
fill!(xduals, 0)
ode_cache_jac = SciMLBase.__init(remake(prob, u0 = xduals, tspan = eltype(xduals).(prob.tspan)), Tsit5())
function loss_jac(du, u)
    SciMLBase.reinit!(ode_cache_jac, u)
    sol = solve!(ode_cache_jac)
    du .= sol[end]
end

jac(J, y, p) = DI.jacobian!(loss_jac, y, J, jac_cache, backend, y)
A = zeros(Float64, 3, 3)
b = [1.0, 2.0, 3.0]
jac(A, b, nothing)
```

lib/BoundaryValueDiffEqFIRK/src/firk.jl
ErikQQY: Fixed

lib/BoundaryValueDiffEqShooting/src/multiple_shooting.jl
gdalle: I think I understand the issue. What would be the right abstraction to add to DI here? Something like `DI.overloaded_inputs` for operator-overloading backends?

ErikQQY: Sorry, I don't fully understand that adding operator-overloading in DI has something to do here, could you please elaborate on this? The Dual numbers are required for the initialization of the loss function so that we can build Jacobian no matter whether we are using `AutoForwardDiff` or `AutoSparse(AutoForwardDiff)`, that's why we use `jac_cache.config.duals` and `jac_cache.pushforward_prep.xdual_tmp` to obtain the Dual type and enable AD. So I wonder if it's possible to have an easy way to get the Dual number in Jacobian preparation?

gdalle: I understand, and I'm looking for a way to put this in DI's API so that it becomes more than a ForwardDiff-specific functionality. For most AD backends based on operator overloading, the functions are called on a different input type (like Dual), so I thought returning that input type could make sense in general

gdalle: I have opened https://github.com/JuliaDiff/DifferentiationInterface.jl/issues/668 to keep track

gdalle: The feature is available in DI v0.6.30 but it is not public API yet. Let's test it, see how it works and what more we need, then we can make it permanent

ChrisRackauckas:
Rebase?

ErikQQY:
Will open a new PR for this

PR #259, opened 2024-11-27
added support of complex numbers for mesh adaptivity in MIRK method

JulianUlrich:
## Checklist

- [ ] Appropriate tests were added 
- [ ] Any code changes were done in a way that does not break public API
- [ ] All documentation related to code changes were updated
  [COLPRAC](https://github.com/SciML/COLPRAC).
- [ ] Any new documentation only uses public API
  
## Additional context

Complex states seemed not to be compatible with mesh adaptivity. I just took the magnitude of the values in cache instead of their complex representation. Maybe not the best way, but it works...

### DIFF
--- a/lib/BoundaryValueDiffEqMIRK/src/adaptivity.jl
+++ b/lib/BoundaryValueDiffEqMIRK/src/adaptivity.jl
@@ -31,8 +31,8 @@ Generate new mesh based on the defect.
     (abstol, _, _), kwargs = __split_mirk_kwargs(; cache.kwargs...)
     N = length(mesh)
 
-    safety_factor = T(1.3)
-    ρ = T(1.0) # Set rho=1 means mesh distribution will take place everytime.
+    safety_factor = abs.(T(1.3))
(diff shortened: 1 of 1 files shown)


ErikQQY:
@JulianUlrich Thank you for your contribution! I have never heard of using complex numbers in adaptivity, under what circumstances do you want to use this? The added operations in this PR seem unnecessary.

JulianUlrich:
I got an error from adaptivity when having complex states in the model. This can be useful in electrical engineering. F.e. for evaluating the impedance of a transmission line model. Mostly, only the boundary values are needed as a result, so it makes sense to start with a coarse mesh.

ErikQQY:
Yeah, I can understand using complex states when we have impedances in the circuit, but I suspect the current implementation can support such models, can you please share an MWE or some simple models that can illustrate the basic idea, there may be more things to do to support these cases.

ErikQQY:
Yeah, the main issue here is the adaptivity routines, could you please add this as a test case and also add the changes to the FIRK methods adaptivity as well?

PR #261, opened 2024-12-16
Use solution object in all solvers

ErikQQY:
Refactored all boundary conditions' definitions to use the solution object. So this is definitely a breaking release, all downstream packages utilizing BoundaryValueDiffEq.jl(if not using Shooting methods) would be affected.

Some affected packages(some from JuliaHub direct dependents):

- [ ] BVProblemLibrary.jl
- [ ] SciMLBenchmarks.jl
- [ ] DiffEqDocs.jl
- [ ] DifferentialEquations.jl
- [ ] Manifolds.jl
- [ ] QuanEstimation.jl
- [ ] InformationGeometry.jl


lib/BoundaryValueDiffEqFIRK/test/expanded/firk_basic_tests.jl
```diff
@@ -39,10 +39,10 @@ end
 f2(u, p, t) = [u[2], -u[1]]
 
 function boundary!(residual, u, p, t)
-    residual[1] = u[:, 1][1] - 5
-    residual[2] = u[:, end][1]
+    residual[1] = u(0.0)[1] - 5
+    residual[2] = u(5.0)[1]
```
ChrisRackauckas, line 43: start and end should be fine just as array indexing?

ErikQQY: Current syntax only supports start and end like `u(0.0)`, we can support array indexing by extending the functionality of the intermediate solution

ErikQQY: Now the boundary conditions support both interpolation and array indexing.

ChrisRackauckas: Why wouldn't array indexing be supported? The solution objects have to support that as part of the solution interface already?

ErikQQY: The current workaround is to build a solution with `EvalSol`, and we don't have to use a real `ODESolution` here to make it work, which would introduce other unnecessary stuff when building a solution using SciMLBase/DiffEqBase, so both interpolation and indexing methods are implemented on `EvalSol` to keep it simple and configurable(for example, we can directly use a shortcut to evaluate the boundary conditions without using interpolation when the boundary is at start and end)

ChrisRackauckas:
Just missing a test for the array indexing.

lib/BoundaryValueDiffEqFIRK/test/expanded/firk_basic_tests.jl
ChrisRackauckas: I see.

ChrisRackauckas: > we can directly use a shortcut to evaluate the boundary conditions without using interpolation when the boundary is at start and end)

The interpolation should already be able to do that?

PR #267, opened 2025-01-17
Clean the using in each subpackage

ErikQQY:
The current way of using packages is messy, so this PR cleans these files and did some refactoring


ChrisRackauckas (approved):

ErikQQY:
The failing BVPM2 wrapper tests are commented out to not impediment the other tests, and the wrapper solvers will be switched to [BVPInterface.jl](https://github.com/ErikQQY/BVPInterface.jl)(specialized interface package for Fortran BVP solvers) once it's registered

PR #272, opened 2025-01-27
Change ODEIntreface to BVPInterface in extension

ErikQQY:
Trying to address #209

Basically, BVPInterface is the predecessor of ODEInterface with several bug fixes, has more algorithms for BVP solving, and will be interfacing more solvers in the future


ChrisRackauckas:
Amazing!

PR #283, opened 2025-02-13
Use DifferentiationInterface for Jacobian stuff

ErikQQY:
Continue #258


ErikQQY:
With the previous progress in #258, this PR should be ready now, ping @gdalle and @adrhill for some professional advice

test/misc/bigfloat_test.jl
```diff
@@ -39,6 +39,6 @@
     u0 = BigFloat.([1.0, 1.0, 1.0])
     tspan = (0.0, 1.0)
     prob = SecondOrderBVProblem(second_f!, second_bc!, u0, tspan)
-    sol4 = solve(prob, MIRKN4(), dt = 0.01)
-    @test SciMLBase.successful_retcode(sol4.retcode)
+    @test_broken sol4 = solve(prob, MIRKN4(), dt = 0.01)
+    @test_broken SciMLBase.successful_retcode(sol4.retcode)
```
ChrisRackauckas, line 43: What happened with bigfloat support?

adrhill:
I have no complaints besides the overriding of the detector in `AutoSparse` (see https://github.com/SciML/BoundaryValueDiffEq.jl/pull/258#issuecomment-2520457418).

lib/BoundaryValueDiffEqAscher/Project.toml
```diff
@@ -55,7 +58,8 @@ Reexport = "1.2"
 SciMLBase = "2.71"
 Setfield = "1.1.1"
 SparseArrays = "1.10"
-SparseDiffTools = "2.23"
+SparseConnectivityTracer = "0.6.9"
+SparseMatrixColorings = "0.4.10"
```
gdalle, line 62:
```suggestion
SparseMatrixColorings = "0.4.13"
```

lib/BoundaryValueDiffEqAscher/Project.toml
```diff
@@ -55,7 +58,8 @@ Reexport = "1.2"
 SciMLBase = "2.71"
 Setfield = "1.1.1"
 SparseArrays = "1.10"
-SparseDiffTools = "2.23"
+SparseConnectivityTracer = "0.6.9"
```
gdalle, line 61:
```suggestion
SparseConnectivityTracer = "0.6.13"
```

lib/BoundaryValueDiffEqAscher/Project.toml
```diff
@@ -37,13 +39,14 @@ BoundaryValueDiffEqCore = "1"
 ConcreteStructs = "0.2.3"
 DiffEqBase = "6.158.3"
 DiffEqDevTools = "2.44"
+DifferentiationInterface = "0.6.30"
```
gdalle, line 42:
```suggestion
DifferentiationInterface = "0.6.41"
```

lib/BoundaryValueDiffEqFIRK/src/firk.jl
```diff
@@ -444,43 +444,66 @@ function __construct_nlproblem(
     L = length(resid_bc)
     resid_collocation = __similar(y, cache.M * (N - 1) * (stage + 1))
 
-    loss_bcₚ = (iip ? __Fix3 : Base.Fix2)(loss_bc, cache.p)
-    loss_collocationₚ = (iip ? __Fix3 : Base.Fix2)(loss_collocation, cache.p)
+    bc_diffmode = if jac_alg.bc_diffmode isa AutoSparse
+        AutoSparse(get_dense_ad(jac_alg.bc_diffmode);
+            sparsity_detector = SparseConnectivityTracer.TracerLocalSparsityDetector(),
```
gdalle, line 449: are we sure that local sparsity detection is what we want here?

lib/BoundaryValueDiffEqFIRK/src/firk.jl
```diff
@@ -514,40 +535,54 @@ function __construct_nlproblem(
...
+            cache.M, N, jac_alg.diffmode)
+        partition = ifelse(
+            ADTypes.mode(jac_alg.bc_diffmode) isa ADTypes.ReverseMode, :row, :column)
+        constant_matrix_coloring = ifelse(
+            ADTypes.mode(jac_alg.nonbc_diffmode) isa ADTypes.ReverseMode,
+            row_colors, column_colors)(colored_result)
+        AutoSparse(get_dense_ad(jac_alg.diffmode);
+            sparsity_detector = ADTypes.KnownJacobianSparsityDetector(sparsity_pattern(colored_result)),
+            coloring_algorithm = ConstantColoringAlgorithm{partition}(
+                sparsity_pattern(colored_result), constant_matrix_coloring))
```
gdalle, line 563: I don't understand this pattern of first solving the coloring problem and then using a `ConstantColoringAlgorithm` to hold the result. The coloring problem is only solved once during preparation anyway.

lib/BoundaryValueDiffEqShooting/src/single_shooting.jl
```diff
@@ -135,7 +140,7 @@ end
 
 function __single_shooting_jacobian_ode_cache(
         prob, jac_cache, ::DiffCacheNeeded, u0, ode_alg; kwargs...)
-    cache = jac_cache.cache
+    cache = jac_cache.config
```
gdalle, line 143: This is an implementation detail, you are not supposed to access fields of preparation results directly (remember https://github.com/JuliaDiff/DifferentiationInterface.jl/pull/672)

gdalle:
(the comments on the versions and on the rest apply to all subpackages, I just didn't put them everywhere)

test/misc/bigfloat_test.jl
ErikQQY: It's failing because of the missing dispatches for SparseArrays factorization in LinearSolve.jl, I am digging into what's going wrong since it was doing fine before🤔

lib/BoundaryValueDiffEqFIRK/src/firk.jl
ErikQQY: The sparsity pattern of the boundary conditions is more of a local sparsity pattern since usually the boundary conditions are only evaluated in some fixed points

lib/BoundaryValueDiffEqFIRK/src/firk.jl
ErikQQY: You mean we don't have to use `sparsity_pattern` to get the coloring of the Jacobian since the colorings will be computed in the preparation?

lib/BoundaryValueDiffEqShooting/src/single_shooting.jl
ErikQQY: Oh, yeah, my bad, forgot we can get the overloaded input

lib/BoundaryValueDiffEqFIRK/src/firk.jl
gdalle: Yes, the preparation includes sparsity pattern detection and coloring, and all of this is transparent to the user. To me it seems like you re-create a new backend object containing the fixed result of coloring, instead of just keeping the provided backend and letting coloring do its thing at preparation time under the hood.

ErikQQY: If I understand correctly, since we have the sparsity pattern of our sparse Jacobian, we can just use

```julia
AutoSparse(get_dense_ad(jac_alg.diffmode);
    sparsity_detector = ADTypes.KnownJacobianSparsityDetector(J),
    coloring_algorithm = GreedyColoringAlgorithm())
```
to construct our `AutoSparse` backend and let preparation do their work?

gdalle: Yes

lib/BoundaryValueDiffEqFIRK/src/firk.jl
gdalle: Be careful then, the results of sparse preparation with local sparsity detection can only be reused if you can guarantee that the sparsity pattern doesn't change between calls

lib/BoundaryValueDiffEqFIRK/src/firk.jl
```diff
@@ -444,43 +444,62 @@ function __construct_nlproblem(
     L = length(resid_bc)
     resid_collocation = __similar(y, cache.M * (N - 1) * (stage + 1))
 
-    loss_bcₚ = (iip ? __Fix3 : Base.Fix2)(loss_bc, cache.p)
-    loss_collocationₚ = (iip ? __Fix3 : Base.Fix2)(loss_collocation, cache.p)
+    bc_diffmode = if jac_alg.bc_diffmode isa AutoSparse
+        AutoSparse(get_dense_ad(jac_alg.bc_diffmode);
```
gdalle, line 448: Why do you re-create an `AutoSparse` object instead of working with the one that was provided?

lib/BoundaryValueDiffEqFIRK/src/firk.jl
```diff
@@ -514,40 +531,49 @@ function __construct_nlproblem(
...
-        __sparsity_detection_alg(__generate_sparse_jacobian_prototype(
+        sparse_jacobian_prototype = __generate_sparse_jacobian_prototype(
             cache, cache.problem_type,
             @view(cache.bcresid_prototype[1:prod(cache.resid_size[1])]),
             @view(cache.bcresid_prototype[(prod(cache.resid_size[1]) + 1):end]),
-            cache.M, N))
+            cache.M, N)
+        AutoSparse(get_dense_ad(jac_alg.diffmode);
+            sparsity_detector = ADTypes.KnownJacobianSparsityDetector(sparse_jacobian_prototype),
+            coloring_algorithm = GreedyColoringAlgorithm())
```
gdalle, line 554: Why not keep the provided coloring algorithm?

lib/BoundaryValueDiffEqFIRK/src/firk.jl
```diff
@@ -566,39 +592,60 @@ function __construct_nlproblem(
     L = length(resid_bc)
     resid_collocation = __similar(y, cache.M * (N - 1))
 
-    loss_bcₚ = (iip ? __Fix3 : Base.Fix2)(loss_bc, cache.p)
-    loss_collocationₚ = (iip ? __Fix3 : Base.Fix2)(loss_collocation, cache.p)
+    bc_diffmode = if jac_alg.bc_diffmode isa AutoSparse
+        AutoSparse(get_dense_ad(jac_alg.bc_diffmode);
+            sparsity_detector = SparseConnectivityTracer.TracerLocalSparsityDetector(),
+            coloring_algorithm = GreedyColoringAlgorithm())
```
gdalle, line 598: Same remarks on re-creating an `AutoSparse`

lib/BoundaryValueDiffEqFIRK/src/firk.jl
```diff
@@ -566,39 +592,60 @@ function __construct_nlproblem(
...
             J_full_band = BandedMatrix(Ones{eltype(y)}(L + cache.M * (N - 1), cache.M * N),
                 (L + 1, cache.M + max(cache.M - L, 0)))
-            __sparsity_detection_alg(__generate_sparse_jacobian_prototype(
-                cache, cache.problem_type, y, y, cache.M, N))
+            sparse_jacobian_prototype = __generate_sparse_jacobian_prototype(
+                cache, cache.problem_type, y, y, cache.M, N)
         end
+        AutoSparse(get_dense_ad(jac_alg.nonbc_diffmode);
+            sparsity_detector = ADTypes.KnownJacobianSparsityDetector(sparse_jacobian_prototype),
+            coloring_algorithm = GreedyColoringAlgorithm())
```
gdalle, line 623: Same remarks on re-creating an `AutoSparse`

lib/BoundaryValueDiffEqFIRK/src/firk.jl
```diff
@@ -628,31 +675,41 @@ function __construct_nlproblem(
...
+    diffmode = if jac_alg.diffmode isa AutoSparse
+        sparse_jacobian_prototype = __generate_sparse_jacobian_prototype(
             cache, cache.problem_type,
             @view(cache.bcresid_prototype[1:prod(cache.resid_size[1])]),
             @view(cache.bcresid_prototype[(prod(cache.resid_size[1]) + 1):end]),
-            cache.M, N))
+            cache.M, N)
+        AutoSparse(get_dense_ad(jac_alg.diffmode);
+            sparsity_detector = ADTypes.KnownJacobianSparsityDetector(sparse_jacobian_prototype),
+            coloring_algorithm = GreedyColoringAlgorithm())
```
gdalle, line 690: Same remarks on re-creating an `AutoSparse`

lib/BoundaryValueDiffEqMIRK/Project.toml
```diff
@@ -36,6 +38,7 @@ BoundaryValueDiffEqCore = "1"
 ConcreteStructs = "0.2.3"
 DiffEqBase = "6.158.3"
 DiffEqDevTools = "2.44"
+DifferentiationInterface = "0.6.30"
```
gdalle, line 41: update versions for DI, SCT, SMC

ErikQQY:
@gdalle Bump. If there are no other major issues, then this PR would be good to go

lib/BoundaryValueDiffEqShooting/src/multiple_shooting.jl
```diff
@@ -92,33 +93,34 @@ function __solve_nlproblem!(
...
+            coloring_algorithm = alg.jac_alg.diffmode.coloring_algorithm)
+    else
+        alg.jac_alg.diffmode
+    end
 
     resid_prototype_cached = similar(resid_prototype)
-    jac_cache = sparse_jacobian_cache(
-        alg.jac_alg.diffmode, sd_bvp, nothing, resid_prototype_cached, u_at_nodes)
-    jac_prototype = init_jacobian(jac_cache)
+    jac_cache = DI.prepare_jacobian(nothing, resid_prototype_cached, diffmode, u_at_nodes)
```
gdalle, line 114: This will error in general. What is the rationale?

lib/BoundaryValueDiffEqShooting/src/multiple_shooting.jl
```diff
@@ -209,15 +217,27 @@ function __multiple_shooting_init_odecache(
...
+        ensemblealg, prob, jac_cache, ::DiffCacheNeeded,
+        diffmode, alg, nshoots, u; kwargs...)
+    if diffmode isa AutoSparse
+        T_dual = eltype(overloaded_input_type(jac_cache))
+        xduals = zeros(T_dual, size(u))
+    elseif diffmode isa AutoForwardDiff
+        T_dual = eltype(overloaded_input_type(jac_cache))
+        xduals = zeros(T_dual, size(u))
     else
         xduals = reshape(cache.t[1:length(u)], size(u))
```
gdalle, line 242: Is this line tested? It seems like it would error with the new code, and also it relies on internals

lib/BoundaryValueDiffEqShooting/src/single_shooting.jl
```diff
@@ -34,25 +34,17 @@ function SciMLBase.__solve(prob::BVProblem, alg_::Shooting; odesolve_kwargs = (;
...
-    # NOTE: We pass in a separate Jacobian Function because that allows us to cache the
-    #       the internal ode solve cache. This cache needs to be distinct from the regular
-    #       residual function cache
     jac_cache = if iip
-        sparse_jacobian_cache(alg.jac_alg.diffmode, sd, nothing, y_, vec(u0))
+        DI.prepare_jacobian(nothing, y_, diffmode, vec(u0))
     else
-        sparse_jacobian_cache(alg.jac_alg.diffmode, sd, nothing, vec(u0); fx = y_)
+        DI.prepare_jacobian(nothing, diffmode, vec(u0))
     end
```
gdalle, line 43: Preparation on `nothing` is not valid

lib/BoundaryValueDiffEqShooting/src/single_shooting.jl
```diff
@@ -119,27 +117,30 @@ function __single_shooting_loss(u, p, cache, bc::BC, u0_size, pt) where {BC}
...
+        prob, jac_cache, ::DiffCacheNeeded, diffmode, u0, ode_alg; kwargs...)
+    if diffmode isa AutoSparse
+        T_dual = eltype(overloaded_input_type(jac_cache))
+        xduals = zeros(T_dual, size(u0))
+    elseif diffmode isa AutoForwardDiff
+        T_dual = eltype(overloaded_input_type(jac_cache))
+        xduals = zeros(T_dual, size(u0))
     else
-        xduals = cache.t
+        xduals = reshape(cache.t[1:length(u0)], size(u0))
```
gdalle, line 143: Same remark on the use of `cache.t` (`cache` is no longer defined and `t` is an internal field

gdalle (changes requested):

lib/BoundaryValueDiffEqShooting/src/multiple_shooting.jl
ErikQQY: Actually this line can be safely removed since this dispatch is specifically designed for cases using ForwardDiff

lib/BoundaryValueDiffEqShooting/src/single_shooting.jl
ErikQQY: From testing, preparation on `nothing` is indeed valid, we prepare on `nothing` just to get the Dualized array in the preparation to construct the Jacobian of ODE-solving

lib/BoundaryValueDiffEqShooting/src/multiple_shooting.jl
ErikQQY: > It is working fine because you're using it with backends that don't actually call the function during preparation

Yeah indeed, the current structure of BoundaryValueDiffEq.jl is not generalized enough to support backends like Enzyme, and to support other backends requires additional refactorizations, which exceed what this PR is expected, so let's focus on the ForwardDiff and FiniteDiff for now, I will fix other AD backend supports in following PRs

gdalle: Sounds good, I just wanted to let you know that this wasn't a supported use in general, even though it happens to work here

gdalle (changes requested):

lib/BoundaryValueDiffEqShooting/src/multiple_shooting.jl
```diff
@@ -209,19 +217,28 @@ function __multiple_shooting_init_odecache(
...
-    if cache isa ForwardDiff.JacobianConfig
-        xduals = reshape(cache.duals[2][1:length(u)], size(u))
-    else
-        xduals = reshape(cache.t[1:length(u)], size(u))
+        ensemblealg, prob, jac_cache, ::DiffCacheNeeded,
+        diffmode, alg, nshoots, u; kwargs...)
+    if diffmode isa AutoSparse
+        T_dual = eltype(overloaded_input_type(jac_cache))
+        xduals = zeros(T_dual, size(u))
+    elseif diffmode isa AutoForwardDiff
```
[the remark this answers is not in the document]
ErikQQY: Oh, speaking of which, we can unlock PolyesterForwardDiff for collocation methods(like MIRK and FIRK) via DI, I find it troublesome to do the same for Shooting methods. While PolyesterForwardDiff.jl is also an AD backend that relies on operator overloading(actually relies on ForwardDiff.jl), if it's possible get the overloaded input type for the preparation from `AutoPolyesterForwardDiff`? https://github.com/JuliaDiff/DifferentiationInterface.jl/pull/672 doesn't support such functionality for PolyesterForwardDiff.jl because PolysterForwardDiff.jl doesn't have any interfaces for the Dualized arrays. I can understand if it's not viable and we need to warn about this too.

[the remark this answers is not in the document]
ErikQQY: Okay then, I will warn users about this in the documentation.

docs/src/basics/autodiff.md
```diff
@@ -2,7 +2,7 @@
 
 !!! note
     
-    We support all backends supported by DifferentiationInterface.jl. Please refer to
+    We support ForwardDiff.jl, FiniteDiff.jl and PolyesterForwardDiff.jl(PolyesterForwardDiff only for collocation methods) via DifferentiationInterface.jl. Please refer to
```
gdalle, line 5: My only concern is that this is a breaking change

gdalle (changes requested):

docs/src/basics/autodiff.md
ErikQQY: Actually, it's not, the current BoundaryValueDiffEq.jl only supports ForwardDiff.jl and FiniteDiff.jl(restricted by SparseDiffTools.jl), so this instead could unlock other backends(at least add PolysterForwardDiff support)

gdalle (approved):

lib/BoundaryValueDiffEqShooting/src/multiple_shooting.jl
ChrisRackauckas: Why is this `nothing`? You know the types?

PR #285, opened 2025-02-23
Add improved MIRK6 method

ErikQQY:


ErikQQY:
Since this is just another MIRK method, I will just merge then

Issue #34, opened 2018-02-15
GE control


ChrisRackauckas:
http://cs.smu.ca/tech_reports/txt2012_001.pdf

PR #287, opened 2025-02-23
Add global error control adaptivity

ErikQQY:
Fix: #34 

Basically, there are four different modes for the error control adaptivity in MIRK and FIRK methods:
1. `DefectControl` for the defect control.
2. `GlobalErrorControl` for the global error control. For now, `HOErrorControl` utilizes the solution from a higher-order solver to obtain the global error. `REErrorControl` utilizes the solution from doubled mesh to obtain the global error.
3. `SequentialErrorControl`: First use defect control, then use global error control.
4. `HybridErrorControl`: Use the linear combination of the defect and global error.


lib/BoundaryValueDiffEqMIRK/src/adaptivity.jl
```diff
@@ -50,7 +51,116 @@ Generate new mesh based on the defect.
     n_ = T(0.1) * n
     n_predict = ifelse(abs((n_predict - n)) < n_, round(Int, n + n_), n_predict)
 
-    if r₁ ≤ ρ * r₂
+    if r₁ ≤ ρ * r₃
```
ErikQQY, line 54: I don't know where this comes from, but this is basically wrong, and made mesh refinement always halve mesh(because that redistribution clause never gets triggered!!!)

ErikQQY:
I believe this PR is ready now

test/runtests.jl
```diff
@@ -1,22 +1,7 @@
...
-const RETESTITEMS_NWORKER_THREADS = parse(Int,
-    get(ENV, "RETESTITEMS_NWORKER_THREADS",
-        string(max(Hwloc.num_virtual_cores() ÷ max(RETESTITEMS_NWORKERS, 1), 1))))
-
-@info "Running tests for group: $(GROUP) with $(RETESTITEMS_NWORKERS) workers"
-
-ReTestItems.runtests(
-    BoundaryValueDiffEq; tags = (GROUP == "all" ? nothing : [Symbol(GROUP)]),
-    nworkers = RETESTITEMS_NWORKERS, nworker_threads = RETESTITEMS_NWORKER_THREADS)
-=#
```
ChrisRackauckas, line 19: ?

ErikQQY: Shouldn't delete them, added them back

Issue #284, opened 2025-02-14
Is BoundaryValueDiffEqFIRK compatible with the Integrator Interface, or some other monitoring and diagnostic tool?


caMi11er:
The title really says it all.  

In other words, is there some diagnostic tool or procedure to monitor the progress and performance of the fully implcit FIRK solver RadauIIa5, in order to learn how to optimally specify the problem for efficient solution? 

One might suspect that colocation methods are more complex and difficult to monitor.  

thanks

ChrisRackauckas:
We can make one. We really should.

We have been wanting to make a more flexible verbose system, where `verbose` can be a type to turn on/off specific printing and logging. That would be what's needed here.

ErikQQY:
The diagnostic tools in BoundaryValueDiffEq.jl would be easy to add, I will implement them soon

PR #290, opened 2025-03-04
Add verbose for collocation methods

ErikQQY:
Fix: #284


ChrisRackauckas:
`true` isn't quite good enough for what we need. We need more verbosity levels. What we need is in SciMLBase and AbstractVerbosity with BVPVerbosity being an instantiation struct, where `BVPVerbosity(true)` is everything, `BVPVerbosity(false)` is nothing, `BVPVerbosity(nothing)` is the default which has some things true and other things false. Then we should add printing to all sorts of things, and make each one a togglable verbosity switch.

@jClugstor can you write this up to SciMLBase? I think I described this already but don't know where the message went.

Issue #288, opened 2025-03-01
Support more AD backends


ErikQQY:
With #283 merged, the limitation of SparseDiffTools.jl is gone and BoundaryValueDiffEq.jl can support ForwardDiff.jl, FiniteDiff.jl, and PolyesterForwardDiff.jl now! But to support more backends like Enzyme and Mooncake, though DifferentiationInterface makes it look easy, the current implementation still needs to be refactored to a more generalized style to realize that, so this issue is to track the refactoring process so that more AD backends just plug in.

- [x] Enzyme
- [x] Mooncake

ChrisRackauckas:
Does this need to be more precise? There's two parts: the collocation and the constraints. I thought you had Enzyme working on the constraints with forward diff working on the collocation?

ErikQQY:
To have simple collocation methods(without cache and preallocations stuff) working with Enzyme and ForwardDiff is easy, and from testing on the simple implementation(Enzyme reverse mode on constraints+ForwardDiff on collocation) does give a performance boost, but the current implementation(mostly the DiffCache stuff I believe) is only compatible with ForwardDiff and not compatible with Enzyme, so we will need some refactor to a more generalized style so that all AD backend can just plug in.

PR #291, opened 2025-03-19
Support Enzyme and Mooncake in AD backends

ErikQQY:
Part of #288


ErikQQY:
This is just delightful😄

image

ErikQQY:
Looks good now, just need a patch from [FastAlmostBandedMatrices.jl](https://github.com/SciML/FastAlmostBandedMatrices.jl) for the sparse AD of BC and then we are good to go.

ChrisRackauckas:
Registering

ChrisRackauckas:
🎉

PR #302, opened 2025-04-05
Support ForwardDiff 1.0

ErikQQY:
Let's see if the upgrade breaks anything


devmotion:
Tests (eg of Core) did not run with ForwardDiff 1. ForwardDiff 1 seems untested.

ErikQQY:
Indeed, we first need ForwardDiff 1 on upstream packages like NonlinearSolve in https://github.com/SciML/NonlinearSolve.jl/pull/561

PR #303, opened 2025-04-06
Use sources

ErikQQY:


.github/workflows/CI_BoundaryValueDiffEq.yml
```diff
@@ -61,12 +61,6 @@ jobs:
         run: |
           import Pkg
           Pkg.Registry.update()
-          # Install packages present in subdirectories
-          dev_pks = Pkg.PackageSpec[]
-          for path in ("lib/BoundaryValueDiffEqCore", "lib/BoundaryValueDiffEqMIRK", "lib/BoundaryValueDiffEqMIRKN", "lib/BoundaryValueDiffEqFIRK", "lib/BoundaryValueDiffEqShooting", "lib/BoundaryValueDiffEqAscher")
-              push!(dev_pks, Pkg.PackageSpec(; path))
-          end
-          Pkg.develop(dev_pks)
```
avik-pal, line 69: unfortunately 1.10 doesn't pick up sources. it was added in 1.11

ErikQQY: Yeah, indeed, we should just keep the developing scripts

PR #305, opened 2025-04-11
Allow more nested nonlinear solve kwargs in FIRK

ErikQQY:
Instead of only allowing `abstol = nest_tol` in nested nonlinear solving, we can specify more kwargs like `RadauIIa5(; nested_nlsolve = true, nested_nlsolve_kwargs = (; abstol = 1e-6, reltol = 1e-6)` using options in https://docs.sciml.ai/NonlinearSolve/stable/basics/solve/ with this PR.


ErikQQY:
Alrighty, good to go now. Need to update related docs in DiffEqDocs

PR #307, opened 2025-04-15
More abstraction for cache

ErikQQY:
Make a higher-level abstraction and change the default absolute tolerance to `1e-6`.


ErikQQY:
Seems good to go, benchmarking CI expected to fail since the abstract algorithms changed, will merge this PR when CIs are happy.

ChrisRackauckas (approved):

PR #317, opened 2025-04-30
update to ForwardDiff@1

oscardssmith:
This has all been tested locally and all passes. The diff is bigger than just version bumps because removing OrdinaryDiffEq as a test dep should save CI a bunch of time (and saved me deving ~30 packages locally)


ChrisRackauckas (approved):

ChrisRackauckas:
Release

Issue #319, opened 2025-05-05
Has something changed in MIRK4()?


vyudu:
**Describe the example**

This example used to find a solution and ran very quickly, but now hits the MaxIters: 

**Minimal Reproducible Example 👇**
```julia
function lotka(du, u, p, t)
           x = u[1]
           y = u[2]
           du[1] = p[1] * x - p[2] * x * y
           du[2] = -p[3] * y + p[4] * x * y
end

function bc!(resid, u, p, t)
           resid[1] = u(0.)[1] - 1.
           resid[2] = u(0.)[2] - 2.
end
bvp = BVProblem(lotka, bc!, [1., 2.], (0, 10), [7.5, 4., 8, 5])
solve(bvp, MIRK4(), dt = 0.01)
```

**Error & Stacktrace ⚠️**

```julia
MaxIters
```

**Not Working Environment (please complete the following information):**

```julia
Status `/private/var/folders/_v/q17ll8wn3h5dqnc6ylfjk7nc0000gn/T/jl_sz0fge/Project.toml`
  [764a87c0] BoundaryValueDiffEq v5.16.1
```

PR #320, opened 2025-05-08
Fix incorrect interpolant evaluaiton in MIRK

ErikQQY:
Fix: #319 

The `interp_eval!` directly changed `y₀`, which would affect the interpolant evaluation where the original discrete solution is used for computing the interpolations after the mesh refinement.


ChrisRackauckas:
Wait, this isn't correct though? It should be using the post initialization one?

ErikQQY:
> Wait, this isn't correct though? It should be using the post initialization one?

No, after the nonlinear solving converges, but the defect norm is not satisfying, the interpolant should be evaluated with the discrete solution+k_discrete+k_interp on the new mesh, where the discrete solution here should be the pre-initialization one.

ChrisRackauckas:
Let's talk about this in the call. @AayushSabharwal

AayushSabharwal:
Sure, but I think I'm missing some context here

ErikQQY:
It's the default absolute tolerance changed to `1e-6`, which reveals some adaptivity issues from #319. The BVP stuff in MTK should be fine again with this fix.

PR #329, opened 2025-05-13
Set strict to false in shooting when prepare jacobian

ErikQQY:
Adapt to DifferentiationInterface 0.7.0 breaking change.


ErikQQY:
Tests are using DifferentiationInterface@0.7.0, and CIs are happy about it, let's go!

Issue #282, opened 2025-02-10
Request: `minsol, maxsol` for finding the lowest/highest values of solutions


vyudu:
This would allow MTK to specify boundary constraints that look like `[x(t) < 3]`, where we want every value of the solution to be less than 3. Thinking the codegen would look like 

```julia
function ineqs!(resid, sol, p, t)
    resid[1] = minsol(sol, (0., 1.)) - 3
end
```

vyudu:
@ErikQQY Just for your reference, [this issue](https://github.com/SciML/ModelingToolkit.jl/issues/3378) has the planned interface for MTK's codegen to BVProblem - does this all seem reasonable?

ErikQQY:
I have not tested on a real problem, but since all the `sol` in boundary conditions are solution objects, it seems we can just do `resid[1] = max(sol)-3`?

ChrisRackauckas:
That's not necessarily a good approximation and would converge as O(h) since it's equivalent to the maximum approximation given a linear interpolation. For a BVP intermediate solution, we don't just have a solution at points but a whole collocation polynomial. What you'd want to do is differentiate that polynomial w.r.t. `t` and solve for the roots. This can be done analytically ahead of time and stored. That would give the critical points. The maximum and minimum of the continuous solution is then O(h^n) approximated by the polynomial at the critical point. So then `minsol(sol, (0., 1.))` would evaluate the interpolation at every interval at the critical points, and then take the min/max of the evaluated points. 

We can start with just the discrete min/max as a starting point, but there are lots of scenarios where the latter would be much more accurate. And since this would be used for constraints, the more accurate form would likely be highly preferred down the line.

PR #332, opened 2025-05-17
Add maxsol and minsol

ErikQQY:
Fix: #282 

This PR implements the `maxsol` and `minsol` functionalities using root finding to find the critical points together with the maximum/minimum defined on intermediate solutions. Solely depending on root finding is insufficient to find the maximum and minimum values, especially in scenarios where the solutions are monotonic, so we need the maximum/minimum to cover these cases.


docs/src/tutorials/inequality.md
```diff
@@ -0,0 +1,29 @@
+# Solve BVP with Inequality Boundary Conditions
+
+When dealing with scenarios that boundary conditions are stronger than just explicit values at specific points—such as inequality boundary conditions, where the solution must satisfy constraints like staying within upper and lower bounds—we need a more flexible approach. In such cases, we can use the `maxsol` and `minsol` functions provided by BoundaryValueDiffEq.jl to specify such inequality conditions.
```
ChrisRackauckas, line 3: This case isn't doing inequalities, it's making the maximum equal to a value.

ErikQQY: Yeah, thanks for the elaboration! That makes sense, maxsol/minsol puts more constraints than inequality conditions here, and maxsol/minsol should only focus on the maximum/minimum value satisfying a specific value.

ChrisRackauckas (approved):

Issue #341, opened 2025-06-11
Update to 5.18.0 results in Differentiation Interface Preparation Mismatch Error


JBMcVey:
MRE here is the standard BVP tutorial problem Example 1, using the TwoPointBVProblem problem setup. Execution using BoundaryValueDiffEq.jl version 5.18.0 now produces a Differentiation Interface Mismatch Error.

```julia
using BoundaryValueDiffEq
using Plots
const g = 9.81
L = 1.0
tspan = (0.0, pi / 2.0)
function simplependulum!(du, u, p, t)
    θ = u[1]
    dθ = u[2]
    du[1] = dθ
    du[2] = -(g / L) * sin(θ)
end

function bc2a!(resid_a, u_a, p) # u_a is at the beginning of the time span
    resid_a[1] = u_a[1] + pi / 2 # the solution at the beginning of the time span should be -pi/2
end
function bc2b!(resid_b, u_b, p) # u_b is at the ending of the time span
    resid_b[1] = u_b[1] - pi / 2 # the solution at the end of the time span should be pi/2
end
u0 = [pi / 2.0, pi / 2.0]
bvp2 = TwoPointBVProblem(simplependulum!, (bc2a!, bc2b!), u0, tspan; bcresid_prototype = (zeros(1), zeros(1)))
sol2 = solve(bvp2, MIRK4(), dt = 0.05)
plot(sol2)
```

**Error & Stacktrace ⚠️**

```julia
[...]

PR #342, opened 2025-06-11
Fix preparation mismatch

ErikQQY:
Fix: #341


ChrisRackauckas:
Add the user's test and have it go over a few methods?

ErikQQY:
Sure, it suprises me that the previous CIs did not catch this🤔

ErikQQY:
> Add the user's test and have it go over a few methods?

Actually, I think it is the previous CIs failed to use the correct version, see #343

Issue #336, opened 2025-05-24
Onto DynamicOptimization: Allow for solving via OptimizationProblem and allow inequality constraints


ChrisRackauckas:
Currently the lowering has two paths: if there are more constraint equations than state values then lower to a NonlinearLeastSquaresProblem, otherwise use a NonlinearProblem. But this can be heavily generalized in a few directions to greatly increase the utility of the codebase.

Let's establish some terms here. We have `u' = f(u,p,t)` and `g(u,p) = 0` as the boundary conditions, or more generally the constraint function. `u` is `u(t)` discretized into a collocation vector.

First of all is the "what" is being solved for. Right now it's always to find u(t) given the constraint dynamics. However, one small generalization is to allow for finding [u(t), p], i.e. allow for performing parameter estimation simultaneously to the solving. In theory if you have exactly `length(u0) + length(p)` conditions that can be done via a NonlinearProblem, but that's kind of an odd case. Almost certainly you'd just use `NonlinearLeastSquaresProblem` problem for this to search for a feasible solution. So first generalization is `fit_parameters = false/true`. Note that bvp4c/bvp5c calls this "Unknown Parameters" https://www.mathworks.com/help/matlab/ref/bvp5c.html . There is an issue here I will point out in a second.

Let's call `x = [u]` or `x = [u, p]` the vector of things to solve for given the choice of `fit_parameters`. 

But the next generalization is to allow for using an OptimizationProblem. If you define the optimization problem where `cost(x) = 0` and make your nonlinear constraints for the OptimizationProblem be `[u' - f(u,p,t), g(u,p)]`, then a nonlinear optimization algorithm like IPOPT will solve this constraint satisfaction problem similarly to the NonlinearLeastSquaresProblem. In fact, a Newton method like IPOPT is doing literally the same thing, where it's doing a Newton method to solve for the constraint satisfaction problem, and then it would do interior point stepping to find a lower cost, but since the cost function is a constant it would just take the first satisfactory point it finds. So first things first is to allow for using `OptimizationProblem` in this form and showing that this gives the same result. 

[...]

PR #350, opened 2025-07-14
Optimization based solvers

ErikQQY:
Part of #336 

Some TODOs:

- [x] Multi-points BVP are encountering some DI issues, but two-point BVP are fine, need to dig deeper to see what's going wrong.
- [x] Need some patches in SciMLBase.jl and DiffEqBase.jl for better interface.
- [x] More docstrings and documentations


ErikQQY:
I think this PR is ready to go now.

The optimal control interface still needs some more work considering the complex whole-time interval inequality(similar to `@variable(model, v[1:T] >= 0, start = v_0)`), but let's do that in a following PR since that would involve changing the way we build the loss function.

lib/BoundaryValueDiffEqCore/src/default_internal_solve.jl
```diff
@@ -46,11 +46,37 @@ function __FastShortcutNonlinearPolyalg(::Type{T} = Float64; concrete_jac = noth
...
+end
+@inline function __concrete_solve_algorithm(prob, ::Nothing, ::Nothing)
     if prob isa NonlinearLeastSquaresProblem
         return __FastShortcutBVPCompatibleNLLSPolyalg(eltype(prob.u0))
     else
         return __FastShortcutBVPCompatibleNonlinearPolyalg(eltype(prob.u0))
     end
 end
+
+# Some optimization algorithms (solvers from interfacing packages) don't support the __solve(prob) interface
```
ChrisRackauckas, line 78: Interesting, we should handle this upstream, which ones?

lib/BoundaryValueDiffEqCore/src/utils.jl
```diff
@@ -560,10 +573,14 @@ end
...
 
 # Construct BVP Solution
-function __build_solution(prob::AbstractBVProblem, odesol, nlsol)
+function __build_solution(prob::AbstractBVProblem, odesol, nlsol::SciMLBase.NonlinearSolution)
     retcode = ifelse(SciMLBase.successful_retcode(nlsol), odesol.retcode, nlsol.retcode)
     return SciMLBase.solution_new_original_retcode(odesol, nlsol, retcode, nlsol.resid)
 end
+function __build_solution(prob::AbstractBVProblem, odesol, optsol::SciMLBase.OptimizationSolution)
+    retcode = ifelse(SciMLBase.successful_retcode(optsol), odesol.retcode, optsol.retcode)
+    return SciMLBase.solution_new_original_retcode(odesol, optsol, retcode, zeros(length(first(odesol)))) # Need a patch in SciMLBase
```
ChrisRackauckas, line 582: what needs a patch?

lib/BoundaryValueDiffEqCore/src/utils.jl
```diff
@@ -597,3 +614,111 @@ end
...
+
+@inline function __extract_lcons_ucons(prob::AbstractBVProblem, ::Type{T}, M, N) where {T}
+    lcons = if isnothing(prob.lcons)
+        zeros(T, N*M)
+    else
+        lcons_length = length(prob.lcons)
+        vcat(prob.lcons, zeros(T, N*M - lcons_length))
+    end
+    ucons = if isnothing(prob.ucons)
+        zeros(T, N*M)
```
ChrisRackauckas, line 638: Using zero is likely not good? That will mess with the stability of the solver quite a bit. If inequality constraints are given then the lcons / ucons should be required from the user.

lib/BoundaryValueDiffEqCore/src/utils.jl
```diff
@@ -597,3 +614,111 @@ end
...
+function __construct_internal_problem(prob::AbstractBVProblem, alg, loss, jac,
+        jac_prototype, resid_prototype, y, p, M::Int, N::Int)
+    T = eltype(y)
+    iip = SciMLBase.isinplace(prob)
+    if !isnothing(alg.nlsolve) || (isnothing(alg.nlsolve) && isnothing(alg.optimize))
+        nlf = NonlinearFunction{iip}(loss; jac = jac, resid_prototype = resid_prototype,
+            jac_prototype = jac_prototype)
+        return __internal_nlsolve_problem(prob, resid_prototype, y, nlf, y, p)
+    else
+        optf = OptimizationFunction{true}(__default_cost(prob.f), AutoFiniteDiff(), # Need to investigate the ForwardDiff dual problem
```
ChrisRackauckas, line 661: what problem?

lib/BoundaryValueDiffEqCore/src/utils.jl
```diff
@@ -597,3 +614,111 @@ end
...
+        resid_prototype, y, p, M::Int, N::Int, ::Nothing)
+    T = eltype(y)
+    if !isnothing(alg.nlsolve) || (isnothing(alg.nlsolve) && isnothing(alg.optimize))
+        nlf = NonlinearFunction{true}(loss; jac = jac, resid_prototype = resid_prototype,
+            jac_prototype = jac_prototype)
+        return __internal_nlsolve_problem(prob, resid_prototype, y, nlf, y, p)
+    else
+        optf = OptimizationFunction{true}(
+            __default_cost(prob.f), get_dense_ad(alg.jac_alg.diffmode),
+            cons = loss, cons_j = jac, cons_jac_prototype = jac_prototype)
```
ChrisRackauckas, line 699: It's not easy to see from this, but when doing multiple shooting with this the codegen should be a little different from the nonlinear solve. With multiple shooting you have the endpoint conditions, and you put those into the nonlinear equation to be satisfied. In the optimization, they should be moved to the equality constraints.

PR #369, opened 2025-08-10
Better downgrade CI

ErikQQY:
Remove the old downgrade CIs in test actions, and force the new downgrade CIs use 1.11


ChrisRackauckas:
The CI is working well now. I have to do a few things upstream for it to pass.

PR #380, opened 2025-10-22
Allow more AD backends in optimization based solvers

### LINKED ISSUE #379: Allow more autodiff backends when using internal optimization solvers
Currently, the autodiff backend in internal optimization solver can only be `AutoFiniteDiff()`, because the value update in `EvalSol` after collocation equation evaluation in loss function `__mirk_loss!`, however, such operation prevent proper sparsity detection with SparseConnectivityTracer.jl. We need to restruct the loss funtion being passed to optimization solver so that more autodiff backends can be used.

This issue can affect #374

ErikQQY:
Fix #379 

The previous implementation can't exploit the sparsity pattern of the loss function due to the value updates in loss function, this PR allows sparse AD in optimmization based internal solvers.

**Before this PR**:

image

**With this PR**:

image


ErikQQY:
CI failing seems unrelated

Issue #395, opened 2025-12-20
Scalar differential equations not supported


abhro:
**Describe the bug 🐞**

If f(u, p, t) returns du as a scalar, BoundaryValueDiffEqCore.jl errors out on `__extract_u0` stating "`prob.u0::Float64` is not supported."

**Expected behavior**

General output of a solution (`ODESolution` or something like that)

**Minimal Reproducible Example 👇**

Without MRE, we would only be able to help you to a limited extent, and attention to the issue would be limited. to know more about MRE refer to [wikipedia](https://en.wikipedia.org/wiki/Minimal_reproducible_example) and [stackoverflow](https://stackoverflow.com/help/minimal-reproducible-example).

```julia
using BoundaryValueDiffEq
using OrdinaryDiffEqTsit5: Tsit5

tspan = (0.0, 2.0)

f(u, p, t) = t * u
bca(ua, p) = ua
bcb(ub, p) = ub - 1

u₀ = 1.0

bvproblem = TwoPointBVProblem(f, (bca, bcb), u₀, tspan)
solver = Shooting(Tsit5())
sol = solve(bvproblem, solver; adaptive = false, dt = 0.5)
```

**Error & Stacktrace ⚠️**

```
┌ Warning: Initial guess provided, but will be ignored for Shooting.
└ @ BoundaryValueDiffEqShooting ~/.julia/packages/BoundaryValueDiffEqShooting/tD3HT/src/single_shooting.jl:7
ERROR: LoadError: `prob.u0::Float64` is not supported.
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:44
 [2] __extract_u0
   @ ~/.julia/packages/BoundaryValueDiffEqCore/hFduL/src/utils.jl:483 [inlined]
 [3] __solve(prob::BVProblem{…}, alg_::Shooting{…}; odesolve_kwargs::@NamedTuple{}, nlsolve_kwargs::@NamedTuple{}, verbose::Bool, kwargs::@Kwargs{…})
   @ BoundaryValueDiffEqShooting ~/.julia/packages/BoundaryValueDiffEqShooting/tD3HT/src/single_shooting.jl:8
 [4] solve_call(_prob::BVProblem{…}, args::Shooting{…}; merge_callbacks::Bool, kwargshandle::Nothing, kwargs::@Kwargs{…})
   @ DiffEqBase ~/.julia/packages/DiffEqBase/YLH9r/src/solve.jl:142
[...]

PR #415, opened 2026-01-20
Support scalar BVP

ErikQQY:
Fix: #395 

The Shooting method need some extra work, but this PR works with collocation methods.

Need https://github.com/JuliaDiff/DifferentiationInterface.jl/pull/954


abhro:
Thank you!!!

PR #449, opened 2026-03-05
Fix CIs for Shooting methods

ErikQQY:
It's weird the Shooting CIs are failing


ChrisRackauckas:
> I'd prefer to have test_brokens here. We don't want to remove the tests of things that aren't working.

No, shooting methods need to have a low tolerance in order to have accurate gradients. It's odd that this case worked with default tolerances. For shooting methods we probably want to default the ODE solver to 1e-8, 1e-8 in order to make the gradients smoother.

PR #479, opened 2026-04-24
Fix FIRK v7 precompile: DI strict + VectorOfArray iteration

ChrisRackauckas-Claude:
## Summary

Unblock the `BoundaryValueDiffEqFIRK` precompile workload against `OrdinaryDiffEq.jl` v7 / SciMLBase v3 / RecursiveArrayTools v4. Two distinct bugs fire inside the `@compile_workload` and are both fixed here.

### 1. `DifferentiationInterface.prepare_jacobian` strict-mode mismatch (`lib/BoundaryValueDiffEqFIRK/src/firk.jl`)

In `__perform_firk_iteration` the nonlinear problem is built with `__construct_problem(cache, vec(cache.y₀), copy(cache.y₀))`. Inside `__construct_problem`, jacobian *preparation* is done on `y = copy(cache.y₀)` (a `Base.ReshapedArray` wrapping a `VectorOfArray`), but jacobian *execution* — reached through `NonlinearSolveFirstOrder.GeneralizedFirstOrderAlgorithmCache` → `__firk_mpoint_jacobian!` at `firk.jl:1487` — is handed a plain `Vector{Float64}`. DI's strict check rejects this:

```
PreparationMismatchError (inconsistent types between preparation and execution):
  - x: ❌
    - prep: Base.ReshapedArray{Float64, 1, VectorOfArray{Float64, 2, Vector{Vector{Float64}}}, ...}
    - exec: Vector{Float64}
If you are confident that this check is superfluous, you can disable it by
running preparation with the keyword argument `strict=Val(false)`.
```

Fix: pass `strict = Val(false)` to every `DI.prepare_jacobian` call in `firk.jl`, mirroring what `BoundaryValueDiffEqShooting` already does (`single_shooting.jl:112`/`116`, `multiple_shooting.jl:148`/`240`/`256`).

### 2. `recursive_unflatten!(::AbstractVectorOfArray, ...)` iteration (`lib/BoundaryValueDiffEqCore/src/utils.jl`)

After the DI issue is cleared, precompile still fails with:

```
MethodError: no method matching copyto!(::Float64, ::SubArray{Float64, 1, Vector{Float64}, Tuple{UnitRange{Int64}}, true})
Stacktrace:
 [1] recursive_unflatten!(y::VectorOfArray{Float64, 2, Vector{Vector{Float64}}}, x::Vector{Float64})
   @ BoundaryValueDiffEqCore .../utils.jl:46
 [2] __perform_firk_iteration(...)
   @ BoundaryValueDiffEqFIRK .../firk.jl:562
```

The method body does `for yᵢ in y; copyto!(yᵢ, x[range]); end`, but under RecursiveArrayTools v4 a 2D `VectorOfArray{T,2,...}` is an `AbstractArray` with scalar linear-indexing iteration — so `yᵢ` is a `Float64`, not the inner timestep vector. Iterate `y.u` (the backing `Vector{Vector{Float64}}`) instead, which matches the semantics the rest of this method assumes and the pattern the SciMLBase v3 compat commit (bea82b8e) used elsewhere in the file.

## Test plan

- [x] `BoundaryValueDiffEqFIRK` precompiles on Julia 1.11 with `Pkg.develop` of both fixed libs (115 s).
- [x] `BoundaryValueDiffEqFIRK` precompiles on Julia 1.12 with `Pkg.develop` of both fixed libs (200 s).
- [ ] Upstream CI (runs the full test suite).

Unblocks the `test (DiffEqDevTools, ...)` and `test (DiffEqDevTools_QA, ...)` jobs on SciML/OrdinaryDiffEq.jl#3521.

Co-Authored-By: Chris Rackauckas <accounts@chrisrackauckas.com>


lib/BoundaryValueDiffEqMIRK/test/ensemble_tests.jl
```diff
@@ -12,7 +12,7 @@
         residual[2] = u(1.0)[1]
     end
 
-    prob_func(prob, ctx) = remake(prob, p = [rand()])
+    prob_func(prob, ctx, repeat) = remake(prob, p = [rand()])
```
ChrisRackauckas:
```suggestion
    prob_func(prob, ctx) = remake(prob, p = [rand()])
```

PR #485, opened 2026-05-01
Fix MIRK/FIRK adaptive mesh refinement under RAT v4 (closes #484)

### LINKED ISSUE #484: MIRK 1.15+ adaptive mesh refinement: UndefRefError on nonlinear BVPs
## Summary

`BoundaryValueDiffEqMIRK` 1.15.0 and 1.16.0 throw `UndefRefError` from inside `__perform_mirk_iteration` (mesh refinement code) when solving sufficiently nonlinear BVPs with default kwargs. Setting `adaptive = false` is a clean workaround. MIRK 1.14.0 works correctly on the same problem (with old SciMLBase 2 stack).

## Versions

- `BoundaryValueDiffEqMIRK` v1.15.0, v1.16.0 — affected
- `BoundaryValueDiffEqMIRK` v1.14.0 — not affected
- `BoundaryValueDiffEqCore` 2.5.0
- `SciMLBase` 3.7.1
- `RecursiveArrayTools` 4.3.0
- Julia 1.12.6

## MWE

Standalone — no Manifolds.jl needed. The ODE is the torus geodesic system on a `(R=3, r=2)` torus (smooth and nonlinear enough that the solver decides to refine the mesh).

```julia
using BoundaryValueDiffEqMIRK

const R = 3.0
const r = 2.0

function f!(du, u, p, t)
    sinθ, cosθ = sincos(u[1])
    R_θ = R + r * cosθ
    du[1] = u[3]
    du[2] = u[4]
    du[3] = -u[4]^2 * R_θ * sinθ / r
    du[4] = 2 * r * sinθ / R_θ * u[3] * u[4]
end

a1 = [0.5, -1.2]
a2 = [-0.5, 0.3]

function bc!(residual, u, p, t)
    ua = u(0.0)
    ub = u(1.0)
    residual[1:2] = ua[1:2] - a1
    residual[3:4] = ub[1:2] - a2
end

prob = BVProblem(f!, bc!, vcat(a1, zero(a1)), (0.0, 1.0))
solve(prob, MIRK4(); dt = 0.05)            # ERROR: UndefRefError
solve(prob, MIRK4(); dt = 0.05, adaptive = false)  # Success
```

## Failing trace

```
ERROR: UndefRefError: access to undefined reference
Stacktrace:
  [1] getindex
    @ ./essentials.jl:920 [inlined]
  [2] __perform_mirk_iteration(cache::BoundaryValueDiffEqMIRK.MIRKCache{...})
    @ BoundaryValueDiffEqMIRK ~/.julia/packages/BoundaryValueDiffEqMIRK/dR6jB/src/mirk.jl:342
  [3] solve!(cache::BoundaryValueDiffEqMIRK.MIRKCache{...})
    @ BoundaryValueDiffEqMIRK ~/.julia/packages/BoundaryValueDiffEqMIRK/dR6jB/src/mirk.jl:275
  ...
```

## Likely culprit

`mirk.jl:342` in MIRK 1.16 (https://github.com/SciML/BoundaryValueDiffEq.jl/blob/master/lib/BoundaryValueDiffEqMIRK/src/mirk.jl#L334-L346):

```julia
if info == ReturnCode.Success # Nonlinear Solve Successful and defect norm is acceptable
    if error_norm > abstol
        # We construct a new mesh to equidistribute the defect
        mesh, mesh_dt, _, info = mesh_selector!(cache, controller)
        if info == ReturnCode.Success
            (length(mesh) < length(cache.mesh)) &&
                __resize!(cache.y₀, length(cache.mesh), cache.M)
            for (i, m) in enumerate(cache.mesh)
                interp_eval!(cache.y₀.u[i], cache, m, mesh, mesh_dt)
            end
            __expand_cache!(cache)
        end
    end
end
```

`__resize!` only fires when the new `mesh` is strictly shorter than `cache.mesh`. The subsequent loop iterates `cache.mesh` and writes into `cache.y₀.u[i]`, so it relies on `length(cache.y₀.u) >= length(cache.mesh)`. If `cache.y₀.u` was allocated with placeholder slots that were never filled (the `safe_similar(u0, ifelse(adaptive, N, 0))` path in `__init`, where the inner vectors are length-0), then `cache.y₀.u[i]` may be an undef-ref `Vector{Float64}` and the `getindex` inside `interp_eval!` blows up.

## Impact

This blocks every Manifolds.jl user who hits `solve_chart_log_bvp` / `estimate_distance_from_bvp` on the SciMLBase 3 stack (https://github.com/JuliaManifolds/Manifolds.jl/pull/887). The Manifolds extension currently works around it with `adaptive = false`, but ideally users should be able to opt back into adaptive refinement once this is fixed.

## Workaround

Pass `adaptive = false` to `solve` when using MIRK on a nonlinear BVP that would otherwise trigger refinement.

ChrisRackauckas-Claude:
## Summary

- Closes #484 (`UndefRefError` from MIRK 1.15+ adaptive mesh refinement on nonlinear BVPs).
- Root cause: `__resize!(::AbstractVectorOfArray, n, M)` in `BoundaryValueDiffEqCore/src/utils.jl` (and the FIRK-specific overloads in `BoundaryValueDiffEqFIRK/src/utils.jl`) were missed by the SciMLBase-v3 / RecursiveArrayTools-v4 compat patch (bea82b8). Under RAT v4, `AbstractVectorOfArray` subtypes `AbstractArray` with column-major linear-indexing semantics, so `length(x)` returns the total scalar count and `last(x)` returns the last scalar. The old code used both as if they were the timestep count / last timestep vector, so `N` was miscomputed and the inner `x.u` was never appended to.
- Bumps `BoundaryValueDiffEqCore` patch version to `2.5.1`.
- Adds a regression `@testitem` in `mirk_basic_tests.jl` based on the issue MWE (torus geodesic BVP — smooth and nonlinear enough that the solver decides to refine).

## Diff

```julia
# Before (BoundaryValueDiffEqCore/src/utils.jl)
function __resize!(x::AbstractVectorOfArray, n, M)
    N = n - length(x)                                                                    # ← scalar count under RAT v4
    N == 0 && return x
    N > 0 ? append!(x, VectorOfArray([safe_similar(last(x)) for _ in 1:N])) : resize!(x, n)
                                                #          ^^^^^^^^^^ last scalar under RAT v4
    return x
end

# After
function __resize!(x::AbstractVectorOfArray, n, M)
    N = n - length(x.u)
    N == 0 && return x
    N > 0 ? append!(x.u, [safe_similar(last(x.u)) for _ in 1:N]) : resize!(x, n)
    return x
end
```

Same pattern applied to the FIRK-specific overloads for `FIRKTableau{false}` and `FIRKTableau{true}`. `resize!(x, n)` is left as-is because RAT v4's `Base.resize!(::AbstractVectorOfArray, ::Integer)` already delegates to `resize!(x.u, n)`.

## Local verification

Issue 484 MWE on Julia 1.12.4 with this branch:

```
=== Adaptive (the failing case) ===
retcode: Success  length(sol.u): 41
sol.u[1]: [0.5, -1.2, -0.4865496830944581, 1.5910097544175474]
sol.u[end]: [-0.5, 0.3, -0.4865763588189779, 1.591006957693632]
SUCCESS: adaptive ran without UndefRefError

=== Non-adaptive (the workaround, should still work) ===
retcode: Success  length(sol2.u): 21
SUCCESS: adaptive=false still works
```

## Test plan

- [ ] CI: BoundaryValueDiffEqCore tests pass
- [ ] CI: BoundaryValueDiffEqMIRK tests pass (incl. the new "Adaptive mesh refinement (issue 484)" testitem)
- [ ] CI: BoundaryValueDiffEqFIRK tests pass
- [ ] Manifolds.jl downstream (the original consumer) — they currently work around with `adaptive = false`; once this lands they can drop the workaround.

### DIFF
--- a/lib/BoundaryValueDiffEqMIRK/test/mirk_basic_tests.jl
+++ b/lib/BoundaryValueDiffEqMIRK/test/mirk_basic_tests.jl
@@ -712,3 +712,39 @@ end
     sol2 = solve(bvp2, MIRK4(), dt = 0.1, adaptive = false, nlsolve_kwargs = (; maxiters = 0))
     @test sol2.u == u_guess
 end
+
+# https://github.com/SciML/BoundaryValueDiffEq.jl/issues/484
+# Adaptive mesh refinement on a sufficiently nonlinear BVP must not throw
+# UndefRefError. The torus geodesic system is smooth and nonlinear enough
+# that the solver decides to refine the mesh.
+@testitem "Adaptive mesh refinement (issue 484)" begin
+    using BoundaryValueDiffEqMIRK
+
+    R = 3.0
+    r = 2.0
+
+    function f!(du, u, p, t)
+        sinθ, cosθ = sincos(u[1])
+        R_θ = R + r * cosθ
+        du[1] = u[3]
+        du[2] = u[4]
+        du[3] = -u[4]^2 * R_θ * sinθ / r
+        du[4] = 2 * r * sinθ / R_θ * u[3] * u[4]
+    end
+
+    a1 = [0.5, -1.2]
+    a2 = [-0.5, 0.3]
+
+    function bc!(residual, u, p, t)
+        ua = u(0.0)
+        ub = u(1.0)
+        residual[1:2] = ua[1:2] - a1
+        residual[3:4] = ub[1:2] - a2
+    end
+
+    prob = BVProblem(f!, bc!, vcat(a1, zero(a1)), (0.0, 1.0))
+    sol = solve(prob, MIRK4(); dt = 0.05)
+    @test SciMLBase.successful_retcode(sol)
+    @test sol.u[1][1:2] ≈ a1 atol = 1.0e-8
+    @test sol.u[end][1:2] ≈ a2 atol = 1.0e-8
+end
--- a/lib/BoundaryValueDiffEqMIRKN/Project.toml
+++ b/lib/BoundaryValueDiffEqMIRKN/Project.toml
@@ -1,7 +1,7 @@
 name = "BoundaryValueDiffEqMIRKN"
 uuid = "9255f1d6-53bf-473e-b6bd-23f1ff009da4"
 authors = ["Qingyu Qu <erikqqy123@gmail.com>"]
-version = "1.15.0"
+version = "1.15.1"
 
 [deps]
 ADTypes = "47edcb42-4c32-4615-8424-f2b9edc5f35b"
--- a/lib/BoundaryValueDiffEqMIRKN/src/mirkn.jl
+++ b/lib/BoundaryValueDiffEqMIRKN/src/mirkn.jl
@@ -137,7 +137,7 @@ function SciMLBase.solve!(cache::MIRKNCache{iip, T}) where {iip, T}
 end
 
 function __perform_mirkn_iteration(cache::MIRKNCache)
-    nlprob = __construct_nlproblem(cache, vec(cache.y₀), copy(cache.y₀))
+    nlprob = __construct_nlproblem(cache, copy(vec(cache.y₀)), copy(cache.y₀))
     solve_alg = __concrete_solve_algorithm(nlprob, cache.alg.nlsolve, cache.alg.optimize)
     kwargs = __concrete_kwargs(
         cache.alg.nlsolve, cache.alg.optimize, cache.nlsolve_kwargs, cache.optimize_kwargs,


ChrisRackauckas-Claude:
Fixed three more RAT-v4 iteration semantics misses surfaced by the LTS CI run on the previous commit:

| Location | Pattern | Symptom on CI |
|---|---|---|
| `lib/BoundaryValueDiffEqMIRK/src/adaptivity.jl:39,93,148` | `[maximum(abs, d) for d in errors]` (`cache.errors::VectorOfArray`) | `DimensionMismatch: array could not be broadcast to match destination` at `ŝ ./= mesh_dt` (test "Mesh redistribution in adaptivity / MIRK3") |
| `lib/BoundaryValueDiffEqFIRK/src/adaptivity.jl:263` | same pattern with `defect` | (would surface in any FIRK adaptive run that hits redistribution) |
| `lib/BoundaryValueDiffEqShooting/src/multiple_shooting.jl:531-532` | `length(first(u0_))`, `similar(first(u0_), …)` (`u0_::VectorOfArray`) | `MethodError: no method matching similar(::Float64, ::Int64)` (test "Manifolds.jl Integration / Initial Guess Functions") |

All three are the same class of bug: under RAT v4, iterating / `first` / `last` on an `AbstractVectorOfArray` works on scalars (`prod(size(x))` length) instead of timesteps. Fix is the same surgical pattern as the previous commit — switch to `.u` / `first(x.u)`.

Local re-run of the failing testitem on Julia 1.12.4:

```
--- MIRK2 ---
retcode: Failure  length(sol.u): 801   # convergence-only failure, not a crash
--- MIRK3 ---
retcode: Success  length(sol.u): 1804
--- MIRK4 ---
retcode: Success  length(sol.u): 360
--- MIRK5 ---
retcode: Success  length(sol.u): 703
--- MIRK6 ---
retcode: Success  length(sol.u): 297
```

Bumps MIRK / FIRK / Shooting patch versions to 1.16.1.

ChrisRackauckas-Claude:
1.11/lts CI failure analysis: this isn't a fresh RAT issue — it's a pre-existing type instability that was masked on master.

**What's happening**

`@inferred solve(mpbvp_iip, MIRK2(; jac_alg); dt = 0.2)` produces an inferred type with the inner `original::SciMLBase.NonlinearSolution{T, N, uType, R, P, A, O, uType2, S, Tr}` UnionAll'd over **all** parameters — i.e., inference loses every parameter of the nested NonlinearSolution. The actual return is fully concrete, hence the `@inferred` mismatch. Root cause is the NonlinearSolve polyalg returning one of several `NonlinearSolution{...}` variants and inference giving up rather than computing a tight Union.

**Why master CI passed but this PR fails**

The Lotka-Volterra test problem is right at the boundary of triggering adaptive mesh refinement. On the master CI runner numerics it converges before refinement, so `mesh_selector!` is never reached and `__perform_mirk_iteration` runs exactly once — only one `NonlinearSolution` ever gets constructed at runtime, which apparently lets inference stay tighter for that path. Locally on Julia 1.11.9 (and on the d11a29b CI runner), the same problem **does** trigger refinement, and inference widens. Reproduced locally:

```
=== Test 1: solve runs ===
retcode: Success  type: ODESolution{Float64, 2, Vector{Vector{Float64}}, …, NonlinearSolution{Float64, 1, Vector{Float64}, …}, Nothing}

=== Test 2: @inferred ===
FAIL: ErrorException
return type ODESolution{…, NonlinearSolution{Float64, 1, …}, Nothing}
does not match inferred return type ODESolution{…, NonlinearSolution{T, N, uType, R, P, A, O, uType2, S, Tr}, Nothing} where {…}
```

I confirmed by reverting just the second commit (`d11a29b`) on top of master locally — that variant doesn't reach `@inferred` at all, because `solve` itself `DimensionMismatch`'s in `mesh_selector!`. So nothing in this PR newly broke type inference; the PR just made the path reachable.

**What I did**

`85f7143` comments out both MIRK `@inferred` blocks (Multi-Point and Two-Point) in `test/misc/type_stability_tests.jl`, mirroring the pre-existing Shooting comment-out just above which already calls out "deep type instability from NonlinearSolve/OrdinaryDiffEq that requires further investigation". Same upstream investigation, same workaround. Open to a different approach (`@test_broken`, narrow-scope `@inferred`, etc.) if you'd prefer.

The real fix lives in NonlinearSolve's polyalg machinery — out of scope for this PR.

ChrisRackauckas-Claude:
You were right — the polyalg itself is type-stable post #870. The widening I was seeing is downstream of that, triggered specifically by the **type** of `nlprob.u0`.

**Real root cause:** under RecursiveArrayTools v4, `vec(::VectorOfArray)` returns a `Base.ReshapedArray{T, 1, VectorOfArray{...}, …}` (in v3 it returned a plain `Vector`). NonlinearSolve's polyalg can't propagate `T, N, uType, R` of the resulting `NonlinearSolution` when `u0` is that ReshapedArray-of-VectorOfArray, so the inferred return type widens to `NonlinearSolution{_A, _B, _C, _D, …}`.

Minimal repro, no BVDE involved:

```julia
using NonlinearSolve, RecursiveArrayTools
f!(du, u, p) = (du .= u .^ 2 .- p)
voa = VectorOfArray([[1.0], [1.0], [1.0]])

prob_resh = NonlinearProblem(NonlinearFunction(f!), vec(voa), [2.0,2.0,2.0])
Base.return_types(solve, Tuple{typeof(prob_resh)})
# → NonlinearSolution{_A, _B, _C, _D, NonlinearProblem{ReshapedArray{…}, …}, …}

prob_flat = NonlinearProblem(NonlinearFunction(f!), reduce(vcat, voa.u), [2.0,2.0,2.0])
Base.return_types(solve, Tuple{typeof(prob_flat)})
# → NonlinearSolution{Float64, 1, Vector{Float64}, Vector{Float64}, …}   (concrete)
```

So another RAT-v4 ripple — `vec(VOA)` semantics changed under v4 in a way that's invisible at the call site but breaks polyalg's type inference downstream.

`562c552` flips MIRK / FIRK / MIRKN's `__perform_*_iteration` from

```julia
nlprob = __construct_problem(cache, vec(cache.y₀), copy(cache.y₀))
```

to

```julia
nlprob = __construct_problem(cache, reduce(vcat, cache.y₀.u), copy(cache.y₀))
```

`recursive_unflatten!(cache.y₀, sol_nlprob.u)` afterwards still copies the result back, so the loss of view-aliasing on the (already-internally-copied-by-the-solver) `u0` is benign.

Also reverts the workaround in `test/misc/type_stability_tests.jl` that commented out the MIRK `@inferred solve` blocks. They pass locally now on Julia 1.11.9 with this fix:

```
=== Test 2: @inferred ===
PASS
```

Bumps `BoundaryValueDiffEqMIRKN` patch to `1.15.1`. Worth filing an upstream NonlinearSolve issue for the `ReshapedArray{<:VectorOfArray}` u0 inference cliff — happy to do that as a follow-up if you want.

ChrisRackauckas-Claude:
Rebased onto master. PR #473 landed first and covers most of what was here:

| What I had | What master now has |
|---|---|
| `__resize!(::VectorOfArray, n, M)` body fix | `bd4571b` — caller-side fix (`cache.y₀.u` so the `Vector{<:AbstractArray}` dispatch is hit instead) |
| MIRK adaptivity `for d in errors` → `errors.u` | `705d926` |
| FIRK adaptivity `for d in defect` → `defect.u` | `705d926` |
| Shooting `first(u0_)` → `first(u0_.u)` | `b251ddd` |
| MIRK `vec(cache.y₀)` → flat Vector | `4475c0f` (`copy(vec(cache.y₀))`) |
| FIRK `vec(cache.y₀)` → flat Vector | `4475c0f`-equivalent already in firk.jl |
| MIRK/FIRK/Core/Shooting version bumps | already on master at 2.5.1 / 1.16.1 / 1.16.1 / 1.16.1 |

What's left, now in `9355a23`:

- **MIRKN missed the same `vec(::VectorOfArray)` fix** in `__perform_mirkn_iteration` — same RAT-v4 inference cliff. Mirrored #473's pattern: `vec(cache.y₀)` → `copy(vec(cache.y₀))`. Bumps MIRKN to 1.15.1.
- **Regression `@testitem` for issue 484** in `mirk_basic_tests.jl` using the torus-geodesic BVP, since #473 didn't add a regression test.

Branch is now a single commit ahead of master. CI should be green this time around.

PR #492, opened 2026-05-20
GPU accelerated solvers

ErikQQY:
Some TODOs:

- [ ] Finalize the utimate APIs
- [ ] Support for other FIRK solvers
- [ ] Fix NonlinearSolve compatiability
- [ ] Fix SparseMatrixColorings compatiability
- [ ] Fix LinearSolve compatiability


ErikQQY:
After today's disscusion, instead of making GPU support as extensions, we might just make the whole collocation as a KernelAbstractions.jl kernel, and infer the detailed platform from input or explicitly specified.

PR #495, opened 2026-06-04
ci: route CI through SciML/.github reusable workflows (@v1)

ChrisRackauckas-Claude:
## Centralize CI via SciML/.github reusable workflows (`@v1`)

This routes BoundaryValueDiffEq.jl's bespoke GitHub Actions workflows through the shared reusable workflows in `SciML/.github`, preserving the repo's existing matrices, triggers (`master`, `paths`/`paths-ignore`), concurrency, downstream package list, downgrade skip list, and the disabled state of the sublibrary downgrade job.

### Converted workflows

| File | Was | Now (`SciML/.github/.github/workflows/...@v1`) |
| --- | --- | --- |
| `SublibraryCI.yml` (new) | 6 per-sublibrary `CI_BoundaryValueDiffEq*.yml` (Core, Ascher, FIRK, MIRK, MIRKN, Shooting) | `sublibrary-tests.yml@v1` — auto-discovers `lib/*`, replaces all six |
| `DowngradeSublibraries.yml` | bespoke per-lib downgrade matrix (disabled `if: false`, #417) | `sublibrary-downgrade.yml@v1` (kept disabled; `julia-version: 1.11`, `skip: Pkg,TOML`, `allow-reresolve: false`) |
| `Downgrade.yml` | bespoke main-package downgrade | `downgrade.yml@v1` (`julia-version: 1.11`, `skip: Pkg,TOML`, `allow-reresolve: false`) |
| `Downstream.yml` | bespoke IntegrationTest (ModelingToolkit.jl) | `downstream.yml@v1` (`owner: SciML`, `repo: ModelingToolkit.jl`, `group: All`, `julia-version: 1.11`) |
| `Documentation.yml` | bespoke docs build | `documentation.yml@v1` (`secrets: inherit`) |
| `FormatCheck.yml` | bespoke Runic via `fredrikekre/runic-action` | `runic.yml@v1` |
| `SpellCheck.yml` | bespoke `crate-ci/typos` | `spellcheck.yml@v1` |
| `BenchmarkPR.yml` | bespoke AirspeedVelocity benchpkg flow | `benchmark.yml@v1` (`julia-version: 1.11`) |

### Consolidation of per-sublibrary CI

[...]

Issue #417, opened 2026-01-23
Re-enable downgrade tests after upstream compat fixes


ChrisRackauckas-Claude:
The downgrade tests (both main package and sublibraries) have been disabled because they are waiting on upstream dependency compat fixes.

The downgrade CI workflows (`Downgrade.yml` and `DowngradeSublibraries.yml`) are currently set to `if: false` to skip them.

Once the upstream compat bounds are fixed, these tests should be re-enabled by removing the `if: false` lines.

PR #496, opened 2026-06-04
ci: restore bespoke sublibrary CI to fix FIRK zero-test coverage drop from #495

ChrisRackauckas-Claude:
## Fix CI coverage drop from sublibrary CI consolidation (#495)

The CI-centralization PR #495 (merged into `master`) consolidated the six per-sublibrary `CI_BoundaryValueDiffEq*.yml` workflows into a single `SublibraryCI.yml` that calls `SciML/.github/.github/workflows/sublibrary-tests.yml@v1`. An audit found this **dropped test coverage**, so this PR reverts the unsafe part while keeping every faithful main-package conversion.

### The coverage drop

`sublibrary-tests.yml@v1` reads per-sublibrary group config from `lib/<name>/test/test_groups.toml`, defaulting (when that file is absent) to:
- `Core` → sends `GROUP=<pkg>` (e.g. `BoundaryValueDiffEqFIRK`)
- `QA` → sends `GROUP=<pkg>_QA` (e.g. `BoundaryValueDiffEqFIRK_QA`)

But `lib/BoundaryValueDiffEqFIRK/test/runtests.jl` only runs tests when `GROUP` is `"EXPANDED"` or `"NESTED"`:

```julia
const GROUP = (get(ENV, "GROUP", "All"))
if GROUP == "EXPANDED"  # runs expanded/
if GROUP == "NESTED"    # runs nested/
```

This repo has **no** `lib/<name>/test/test_groups.toml`, so the reusable sent group names FIRK's `runtests.jl` does not recognize. Result: **FIRK ran ZERO test items** under the consolidated workflow. The same hazard applies to any sublibrary whose `runtests.jl` keys off custom group tags. The bespoke `CI_BoundaryValueDiffEqFIRK.yml` ran a real `group: [EXPANDED, NESTED]` matrix; the consolidation silently lost it.

### What this PR does

- **Removes** `.github/workflows/SublibraryCI.yml` (the unsafe `sublibrary-tests.yml@v1` caller).
- **Restores verbatim from the base branch** the six bespoke per-sublibrary workflows, so each sublibrary again runs its real `GROUP`/version/OS matrix:
  - `CI_BoundaryValueDiffEqFIRK.yml` — `group: [EXPANDED, NESTED]` on `1.11` + `lts`
  - `CI_BoundaryValueDiffEqAscher.yml`, `...Core.yml`, `...MIRK.yml`, `...MIRKN.yml`, `...Shooting.yml` — `group: core` on `1.11` + `lts` (each `dev`-ing `lib/BoundaryValueDiffEqCore` where needed)
- **Restores verbatim from base** `DowngradeSublibraries.yml`. Its consolidated form routed through `sublibrary-downgrade.yml@v1`, which relies on the same default-`GROUP` mechanism that mis-handles FIRK's groups; keeping it bespoke avoids the same hazard. It remains disabled (`if: false`, #417) exactly as before.
- **Restores `julia-version: "1.10"`** on the `Documentation.yml` caller. The base docs build pinned `version: '1.10'`; `documentation.yml@v1` defaults to `"1"`, so #495 dropped the pin. This restores it.

### Sublibrary CI intentionally LEFT BESPOKE

Per the SciML centralization guidance ("MONOREPO SUBLIBRARY CONSOLIDATION — DANGER"): do not consolidate per-sublibrary `CI_*.yml` into `sublibrary-tests.yml@v1` unless each sublibrary still runs its real tests. Because the sublibraries' `runtests.jl` files dispatch on custom group names (FIRK: `EXPANDED`/`NESTED`) and there are no `test_groups.toml` files to remap them, the sublibrary CI is **left bespoke** here. This is the conservative, no-coverage-loss choice.

### Faithful main-package conversions KEPT (not flagged by the audit)

These remain centralized as in #495:

| File | Reusable |
| --- | --- |
| `BenchmarkPR.yml` | `benchmark.yml@v1` (`julia-version: 1.11`) |
| `Documentation.yml` | `documentation.yml@v1` (now with `julia-version: 1.10` restored) |
| `Downgrade.yml` | `downgrade.yml@v1` (`julia-version: 1.11`, `skip: Pkg,TOML`, `allow-reresolve: false`) |
| `Downstream.yml` | `downstream.yml@v1` (ModelingToolkit.jl / All, `julia-version: 1.11`) |
| `FormatCheck.yml` | `runic.yml@v1` |
| `SpellCheck.yml` | `spellcheck.yml@v1` |

`CI_BoundaryValueDiffEq.yml` (main-package tests) and `QA.yml` were already left bespoke in #495 and are unchanged here.

### Proper follow-up (separate, maintainer-reviewed)

To eventually centralize the sublibrary CI **without** losing coverage, a maintainer should, as its own reviewed step:
1. Add `lib/<name>/test/test_groups.toml` to each sublibrary encoding the real groups, versions, OS, and threads — e.g. for FIRK something like `[EXPANDED]` and `[NESTED]` entries with the `1.11`/`lts` version matrix.
2. Align each `lib/<name>/test/runtests.jl` so it honors the `GROUP` value the reusable sends (the reusable maps a group named `Core` to `GROUP=<pkg>` and any other group `G` to `GROUP=<pkg>_G`), instead of the current `EXPANDED`/`NESTED` hardcoding.
3. Only then replace the bespoke `CI_*.yml` files with a single `sublibrary-tests.yml@v1` caller, and verify in CI that every group actually executes a non-zero number of test items.

### Validation performed

- `python3 yaml.safe_load` parses all 16 workflow files cleanly.
- Every remaining reusable caller references a real `SciML/.github/.github/workflows/<name>.yml@v1` and passes only inputs that workflow declares.
- Diff vs the pre-#495 base confirms the only workflow files that still differ from base are the six faithful main-package conversions; all sublibrary CI is byte-identical to base.


ChrisRackauckas-Claude:
Superseded by https://github.com/SciML/BoundaryValueDiffEq.jl/pull/497: keeping the centralized sublibrary-tests@v1 form and fixing coverage via test_groups.toml + runtests GROUP parsing, instead of reverting to bespoke CI_*.yml.

PR #523, opened 2026-07-03
Move heavy AD backends (Enzyme/Mooncake) into a group env for FIRK and MIRK

ChrisRackauckas-Claude:
> **Ignore until reviewed by @ChrisRackauckas.** Draft — not ready for merge.

## What

Moves the heavy optional AD backends **Enzyme** and **Mooncake** out of the old-style `[extras]`/`[targets].test` of the `BoundaryValueDiffEqFIRK` and `BoundaryValueDiffEqMIRK` sublibraries and into a dedicated **AD test group** with its own `test/AD/Project.toml`, following the validated recipe from the Integrals.jl (#339) / DataInterpolations.jl (#552) pilots.

## Why

In both sublibraries, Enzyme and Mooncake are listed in `[extras]`+`[targets].test` but **not** in `[weakdeps]` (they are pure test deps, not extension triggers). The sublibrary Downgrade lane (`DowngradeSublibraries.yml` runs each lib's `Core` group at floor via `sublibrary-downgrade.yml@v1`) merges the old-style test extras into `[deps]` as **hard** deps and force-min-resolves the whole graph jointly. Because Enzyme+Mooncake are extra-only, they land in that base joint at-floor resolve — exactly the joint-floor scenario this migration eliminates.

Only the different-AD-backend compatibility tests use these backends:
- FIRK: `test/expanded/ad_tests.jl` (was in the EXPANDED group)
- MIRK: `test/Core/ad_tests.jl` (was in the Core group — so the Downgrade `Core` lane directly resolved them)

## Changes (Project.toml + test wiring only)

**FIRK**
- `git mv test/expanded/ad_tests.jl -> test/AD/ad_tests.jl`
- New `test/AD/Project.toml`: `BoundaryValueDiffEqFIRK` (+ `BoundaryValueDiffEqCore` sibling) via `[sources]`, `Enzyme`, `Mooncake`, `ForwardDiff`, `Test`, `SafeTestsets`, `SciMLTesting`; `[compat]` copies the lib's floors.
- `runtests.jl`: remove `ad_tests` from the EXPANDED body; add an `"AD"` group with `env = joinpath(@__DIR__, "AD")`; add `"AD"` to `all`.
- `test_groups.toml`: declare `[AD]` (matches EXPANDED/NESTED matrix).
- Root `Project.toml`: drop `Enzyme`/`Mooncake` from `[extras]`+`[targets].test` and their `[compat]` (extra-only, not weakdeps).

**MIRK**: same shape (`test/Core/ad_tests.jl -> test/AD/ad_tests.jl`, add `"AD"` group + `test_groups.toml` entry, trim `Project.toml`). `OptimizationIpopt` (a genuine Core test dep) is intentionally kept.

The root package and the other four sublibraries (Ascher, Core, MIRKN, Shooting) have **no** heavy AD backends in their own extras — not applicable, left untouched.

## Validation (local, Julia 1.12)

Downgrade base resolve (jdc-52 harness) for each migrated lib:
- **FIRK BEFORE**: base merged manifest installs full `Mooncake 0.5.15` (+ Enzyme). **AFTER**: only lightweight `EnzymeCore 0.8.21` remains transitively; full Enzyme/Mooncake gone; root `Successfully resolved`.
- **MIRK BEFORE**: base manifest installs full `Enzyme 0.13.100` + `Mooncake 0.5.15`. **AFTER**: only `EnzymeCore 0.8.21` remains; full Enzyme/Mooncake gone.

AD group env instantiates cleanly (develops the sublibrary by path): Enzyme 0.13.173, Mooncake 0.5.36, ForwardDiff 1.4.1, with all AD extensions precompiling. The main/Core test env provably no longer contains Enzyme or Mooncake.


ChrisRackauckas-Claude:
Note on local validation: running the full \`GROUP=Core\` suite for BoundaryValueDiffEqMIRK on Julia 1.12.6 shows **2 failures** — both in the \`OptimizationIpopt\` path of \`test/Core/mirk_basic_tests.jl\`:

- `BVP with inequality constraints`
- `StandardBVProblem optimize path without user lcons/ucons`

These are **pre-existing and unrelated to this PR**. Reproduced byte-identical on a clean checkout of unmodified `master` (commit 49a1718), running just those two `@test_nowarn` solves against the same resolved versions (OptimizationIpopt 1.2.2, OptimizationBase 5.1.4, Ipopt 1.15.0). The Ipopt solve itself succeeds (`EXIT: Optimal Solution Found.`); the `@test_nowarn` fails only because `IpoptOptimizer` now emits an `@info` log that `contains_warn` counts as a warning:

```
Info: Verbosity toggle: unsupported_kwargs
  common abstol is currently not used by IpoptOptimizer(...)
```

This PR does not touch these testsets or their dependencies (it only moves the Enzyme/Mooncake AD tests into a separate group env). The AD group and the rest of the Core suite (198 assertions) pass. Filing a separate note on the OptimizationIpopt verbosity warning.

PR #538, opened 2026-07-10
Document public API coverage

ChrisRackauckas-Claude:
This PR should be ignored until reviewed by @ChrisRackauckas.

## Summary
- add source-site docstrings for public/exported BoundaryValueDiffEqCore API gaps
- add rendered docs entries for exported/public bindings, including core internal interfaces
- add QA coverage that checks exported, native public, and SciMLPublic @public names for source docstrings and docs @docs entries

## Validation
- `Runic.jl --check` on touched Julia files: passed
- `julia --project=test/qa -e "include(\"test/qa/public_api_docs.jl\")"`: passed, 156/156 assertions
- `julia --project=test/qa test/qa/qa.jl`: passed, 17/17 QA checks after the coverage test
- `julia --project=docs docs/make.jl`: attempted locally; fails in pre-existing `docs/src/tutorials/extremum.md` example with `MethodError: no method matching Float64(::ForwardDiff.Dual{...})`
- clean `origin/master` worktree reproduces the same `extremum.md` example failure, so the docs build failure is not introduced by this PR


ChrisRackauckas-Claude:
Validation note after rebasing onto current `origin/master`: `julia --project=docs docs/make.jl` still fails in the pre-existing `docs/src/tutorials/extremum.md` example block with `MethodError: no method matching Float64(::ForwardDiff.Dual{...})`. I reproduced the narrowed example on clean current `origin/master` (`fa36d42a`) and filed it separately as #539.

PR #547, opened 2026-07-10
Avoid FIRK AD residual collection allocation

ChrisRackauckas-Claude:
## What changed

The FIRK `NoDiffCacheNeeded` collocation loss now passes a view of `residual[2:end]` to `Φ!` and flattens that view with a local `__firk_flatten_residuals!` loop, avoiding the GC-sensitive temporary vector at the site where Enzyme forward-mode AD crashed.

The sibling `StandardBVProblem` `NoDiffCacheNeeded` `__firk_loss!` had the identical `[r for r in residual]` collection; it now uses `residual` directly (`Φ!` on `residual[2:end]`, `eval_bc_residual!` on `residual[1]`, `recursive_flatten!` on `residual`), matching the TwoPointBVProblem `NoDiffCacheNeeded` path which already did this.

This is the focused AD portion of https://github.com/SciML/BoundaryValueDiffEq.jl/pull/552, rebased onto current `master` (`8e524b58`).

## Correctness notes

- `Φ!` only mutates the inner residual arrays in place (`f!(residual[i], ...)`, broadcast `.=`), never assigns to the outer container, so a view of the container is semantically identical to the old copy.
- `recursive_flatten!` in BoundaryValueDiffEqCore is restricted to `x::Vector{<:AbstractArray}` and cannot accept the `SubArray` view; the local helper is the same loop with the container argument unconstrained. (Widening the Core signature would work too but requires a Core release before FIRK could use it.)
- `residual` is `Vector{<:AbstractArray}` by construction, so `recursive_flatten!(resid, residual)` in `__firk_loss!` dispatches exactly as the old copied `resids` did.

## Status of the original crash

The motivating clean-master signal-11 was an Enzyme 0.13.182 regression, fixed upstream in 0.13.183 (audit comment below). This PR remains worthwhile as allocation hardening at the exact frame from the crash report — it removes one temporary-vector allocation per loss/jacobian evaluation in both `NoDiffCacheNeeded` paths.

## Validated locally

Julia 1.12.4, resolver-chosen Enzyme 0.13.203 + Mooncake 0.5.57, on the rebased commit:

```text
BOUNDARYVALUEDIFFEQ_TEST_GROUP=AD  FIRK Expanded AD Tests | 9 pass | 9 total | 23m39.6s
BOUNDARYVALUEDIFFEQ_TEST_GROUP=QA  Quality Assurance      | 86 pass | 86 total | 1m36.8s
```

Runic-formatted output is byte-identical to the committed file; `typos` and `git diff --check` clean. Docs build skipped: no docstring or public API change. GPU paths not exercised.

## Follow-up noted (out of scope here)

`lib/BoundaryValueDiffEqMIRK/src/mirk.jl` `__mirk_loss_collocation!` (`NoDiffCacheNeeded`, ~line 601) has the identical `[r for r in residual[2:end]]` pattern on its AD path. Same hardening applies but is kept out of this PR to stay FIRK-scoped.

🤖 Generated with Devin CLI 3000.10.21 (model: swe-2-max; session: local Devin CLI session `polydactyl-gold`, no shareable URL — transcript in `~/.local/share/devin/cli/sessions.db`)

### DIFF
--- a/lib/BoundaryValueDiffEqFIRK/src/firk.jl
+++ b/lib/BoundaryValueDiffEqFIRK/src/firk.jl
@@ -1522,10 +1522,19 @@ end
         resid, u, p, y, mesh, residual, cache, trait::NoDiffCacheNeeded, constraint
     )
     y_ = recursive_unflatten!(y, u)
-    resids = [r for r in residual[2:end]]
+    resids = view(residual, 2:lastindex(residual))
     Φ!(resids, cache, y_, u, trait, constraint)
-    recursive_flatten!(resid, resids)
+    __firk_flatten_residuals!(resid, resids)
     return nothing
+end
+
+@views function __firk_flatten_residuals!(y::AbstractVector, x)
+    i = 0
+    for xᵢ in x
+        copyto!(y[(i + 1):(i + length(xᵢ))], xᵢ)
+        i += length(xᵢ)
+    end
+    return y
 end
 
 @views function __firk_loss_collocation(u, p, y, mesh, residual, cache, trait)


ChrisRackauckas-Claude:
Independent clean-master audit (Julia 1.12.6):

- I ran the native FIRK AD group on `4dfc3ef2872deace50764f53e8727cfbe7623d99` with a fresh local depot and the current resolver. It selected Enzyme 0.13.183 and completed with `FIRK Expanded AD Tests | 9 pass | 9 total | 18m05.5s`; `Pkg.test()` reported `BoundaryValueDiffEqFIRK tests passed`.
- I then pinned the same clean source and Julia to Enzyme 0.13.182 and directly included the unchanged `test/AD/ad_tests.jl`. It reproduced signal 11 at `ad_tests.jl:4` and glibc reported `corrupted double-linked list`. The damaged process did not exit after printing the crash and was terminated after the evidence was captured.
- Hosted history isolates the dependency boundary on the same Julia 1.12.6/source path: [Enzyme 0.13.181 passed 9/9](https://github.com/SciML/BoundaryValueDiffEq.jl/actions/runs/29171577095/job/86594066385), while [0.13.182 failed with signal 11](https://github.com/SciML/BoundaryValueDiffEq.jl/actions/runs/29188568272/job/86639109094). No FIRK source changed across that boundary.
- Enzyme 0.13.183 was released after the failing master job and is what the resolver now selects. The 0.13.182 delta's Julia-1.12-specific root extraction change is [fbbdfab](https://github.com/EnzymeAD/Enzyme.jl/commit/fbbdfabfb6a8b20f34e17c3e5cd5d06627467a9e); that is the leading commit-level cause from source inspection, but I did not claim a commit-level local bisect.

Thus the original clean-master CI crash is an Enzyme 0.13.182 regression that no longer reproduces with the current 0.13.183 resolver. This PR still provides allocation hardening at the exact FIRK frame from the crash, but a duplicate BoundaryValueDiffEq PR is not warranted for restoring CI. The audit clone is clean and no test files or manifests remain.

PR #549, opened 2026-07-10
Fix FIRK vector-of-vector residual AD storage

ChrisRackauckas-Claude:
Ignore this PR until reviewed by @ChrisRackauckas.

## Summary
- Avoid reusing the Float64 `EvalSol` captured during FIRK nonlinear problem construction when evaluating StandardBVProblem boundary residuals.
- Build `EvalSol` from the current unflattened nonlinear state so ForwardDiff dual values stay in AD-compatible storage.
- This fixes the FIRK vector-of-vector initial guess path exposed after splitting the long EXPANDED test group.

## Master Reproduction
On the sharded local reproduction branch, the existing vector-of-vector test failed before this source patch with:

```julia
BOUNDARYVALUEDIFFEQ_TEST_GROUP=EXPANDED_VECTOR julia +1 --startup-file=no --project=lib/BoundaryValueDiffEqFIRK -e 'using Pkg; Pkg.test()'
```

Result: `MethodError: no method matching Float64(::ForwardDiff.Dual...)` at `lib/BoundaryValueDiffEqFIRK/test/expanded/vectorofvector_initials_tests.jl:71`, with FIRK frames `firk.jl:1403`, `firk.jl:677`, and `firk.jl:579`.

## Validation
On this branch:

```julia
timeout 3600 julia +1 --startup-file=no -e 'import Pkg; Pkg.activate(; temp=true); Pkg.develop(Pkg.PackageSpec(path="lib/BoundaryValueDiffEqFIRK")); Pkg.add("OrdinaryDiffEqRosenbrock"); include("lib/BoundaryValueDiffEqFIRK/test/expanded/vectorofvector_initials_tests.jl"); include("lib/BoundaryValueDiffEqFIRK/test/nested/vectorofvector_initials_tests.jl")'
```

passed:
- Expanded `VectorOfVector Initial Condition | 2 / 2` in 59.1s
- Nested `VectorOfVector Initial Condition | 2 / 2` in 1m00.7s

Additional checks:
- `julia +1 --startup-file=no --project=/home/crackauc/sandbox/tmp_20260708_121627_10236/.runic_env -m Runic --check lib/BoundaryValueDiffEqFIRK/src/firk.jl`
- `typos lib/BoundaryValueDiffEqFIRK/src/firk.jl`
- `git diff --check`


ChrisRackauckas-Claude:
Rebased onto current master (`4bfb01fb`, post-#583/#584) and re-validated. This fix is still correct and still needed; adding the trigger analysis, which explains why this failure comes and goes.

## Why it only fires sometimes

The broken write needs something to take a directional derivative through the **combined** residual `__firk_loss!`. The separated bc-loss paths (`firk.jl:1497`, `firk.jl:1507`) already rebuild `EvalSol` from `y_` each call, so they were never affected — which is why ordinary Jacobian assembly does not trip it.

A line search is what takes that directional derivative. Isolating the nonlinear solver on the failing problem, against unpatched master:

| `nlsolve` | unpatched master | with this PR |
| --- | --- | --- |
| default (3-alg polyalgorithm) | **throws** `MethodError: Float64(::Dual)` | Success |
| `NewtonRaphson()` (no line search) | Success | Success |
| `NewtonRaphson(; linesearch = BackTracking())` | **throws** | Success |
| `TrustRegion()` | Success | Success |

The default `__FastShortcutBVPCompatibleNonlinearPolyalg` has `NewtonRaphson(; linesearch = BackTracking())` as its **second** algorithm. That is how the default path reaches the broken code, and it means whether the bug fires depends on whether the first algorithm converges first. An unrelated change in NonlinearSolve or the AD stack can therefore expose or hide this, which matches the on-and-off history.

The stack confirms the route: `LineSearch/backtracking.jl:96` -> `SciMLJacobianOperators` JVP -> `__firk_loss!` at `firk.jl:1403`.

## Validation on current master

Rebased branch (`288ea4e7`), Julia 1.12.6. Full test group, not just the single file:

```
BOUNDARYVALUEDIFFEQ_TEST_GROUP=EXPANDED_MISC julia +1.12 --project=lib/BoundaryValueDiffEqFIRK -e 'using Pkg; Pkg.test()'
```

```
exit 0
FIRK Expanded Ensemble Tests                |  28  28   9m02.1s
FIRK Expanded Singular BVP Tests            |  28  28   9m20.9s
FIRK Expanded VectorOfVector Initials Tests |   2   2   2m14.2s
```

That group errored on unpatched master and now passes; the vectorofvector testset goes from 1 error to 2/2. All four solver configurations in the table above were also re-run against the patched source and succeed.

Runic clean on `lib/BoundaryValueDiffEqFIRK/src/firk.jl`. Rebase was a clean replay of both commits with no conflicts.

## Note on the approach

The `eltype(first(y)) <: eltype(first(eval_sol.u))` guard keeps the preallocated buffer on the non-AD path and only allocates a fresh `EvalSol` when the state carries `Dual`s, so it avoids a per-residual-evaluation allocation in the common case. That is a better trade than unconditionally rebuilding, which is what I would otherwise have proposed in #585.

#585 is a duplicate of this PR — I filed it before finding this — and I am closing it in favour of this one.

ChrisRackauckas-Claude:
CI on the rebased commit (`288ea4e7`) confirms the fix. Marked ready for review.

**`EXPANDED_MISC` passes on all three Julia versions** — that is the group this PR fixes, and it errored on unpatched master.

Also green: `EXPANDED_AFFINENESS`, `EXPANDED_CONVERGENCE`, `EXPANDED_NLLS`, `NESTED_BASIC`, `NESTED_CONVERGENCE`, `NESTED_MISC`, `NESTED_AFFINENESS`, `NESTED_NLLS` (lts/pre), `AD` (lts), `QA`.

Two failures, neither from this change:

- **`AD` / pre** — #554, Enzyme on prerelease.
- **`EXPANDED_BASIC` / 1 and / lts** — the runner-kill flake, not a test failure. Both have `Run tests` at conclusion `null` rather than `failure`, no steps after it, and logs returning `BlobNotFound`. The same signature hit `EXPANDED_BASIC` on the #583 run (lts) and the #584 run (Julia 1), both of which predate this patch, so it is not introduced here.

On that second one: it is now 4 failures across 9 job-runs of `EXPANDED_BASIC`, always with the affected version shifting. Worth noting that this PR's fresh-`EvalSol` path allocates under AD, so in principle it could add memory pressure to the group that is already closest to the runner ceiling (15.4 GB RSS locally against 16 GB) — but the flake predates the patch and `EXPANDED_BASIC` does not exercise the vector-of-vector path, so I do not think they are connected. Filing it separately rather than leaving it implied here.

ChrisRackauckas-Claude:
Correcting my previous comment on two counts.

**1. It was 3/3, not 2/3.** All three `EXPANDED_BASIC` jobs failed on this run (`pre` had not reported when I wrote that). Previous runs failed 1-of-3, so calling it the same flake and moving on was not justified at the time.

**2. I have now measured it, and this PR is not the cause.** Peak RSS for `EXPANDED_BASIC` via `/usr/bin/time -v`, patched tree vs current master, Julia 1.12.6:

| | Peak RSS | Result | Time |
| --- | --- | --- | --- |
| this PR | 16.77 GB (17584508 KB) | exit 0, 126/126 | 49m41s |
| master | 16.82 GB (17633328 KB) | exit 0, 126/126 | 47m07s |

The patched tree uses marginally *less* memory and passes identically, so the fresh-`EvalSol` allocation does not add measurable pressure. The wall-time difference is run-to-run variation; the two ran concurrently on the same machine.

**The underlying problem is that `EXPANDED_BASIC` needs ~16.8 GB peak against a 16 GB standard runner.** That is why the kill always lands on this group, always leaves `Run tests` at conclusion `null` with no retrievable log, and moves between Julia versions. It is a pre-existing condition of the group, not of this change — filed separately as #586.

The fix itself remains confirmed: `EXPANDED_MISC` green on all three versions, where it errored on unpatched master.

PR #563, opened 2026-07-15
Target ModelingToolkitBase in downstream CI

ChrisRackauckas-Claude:
## Summary

- point the ModelingToolkit downstream row at `lib/ModelingToolkitBase`, where the relevant boundary-value problem tests now live
- run the `InterfaceII` group containing `test/bvproblem.jl`
- develop the local `BoundaryValueDiffEqCore`, `BoundaryValueDiffEqMIRK`, and `BoundaryValueDiffEqAscher` projects into that target environment so monorepo changes are actually exercised

This caller depends on SciML/.github#115 being merged and the shared `v1` tag being advanced, because it uses the new `upstream-subdirs` input.

## Local verification

- `actionlint .github/workflows/Downstream.yml` — passed
- `julia +1.12 --startup-file=no -m Runic --check .` — passed
- `git diff --check` — passed
- exact Julia 1.11 rehearsal of the proposed `InterfaceII` target with all three source subprojects developed — the target manifest contained path entries for all three local packages and the intended suite ran for 60m29s: **7,233 passed, 3 errored, 4 pre-existing broken**

The three errors are real MIRK/optimization compatibility failures in the existing source (`Lotka-Volterra`, cost-function compilation, and parameter estimation), not resolver handling or a silent no-op. They are being investigated independently on a clean `master` checkout; this focused workflow PR does not hide or weaken them.

**Please ignore this draft until it has been reviewed by @ChrisRackauckas.**

### DIFF
--- a/.github/workflows/Downstream.yml
+++ b/.github/workflows/Downstream.yml
@@ -17,11 +17,13 @@ jobs:
       fail-fast: false
       matrix:
         package:
-          - {user: SciML, repo: ModelingToolkit.jl, group: All}
+          - {user: SciML, repo: ModelingToolkit.jl, group: InterfaceII, subdir: lib/ModelingToolkitBase, upstream_subdirs: "lib/BoundaryValueDiffEqCore,lib/BoundaryValueDiffEqMIRK,lib/BoundaryValueDiffEqAscher"}
     uses: "SciML/.github/.github/workflows/downstream.yml@v1"
     with:
       owner: "${{ matrix.package.user }}"
       repo: "${{ matrix.package.repo }}"
+      subdir: "${{ matrix.package.subdir }}"
+      upstream-subdirs: "${{ matrix.package.upstream_subdirs }}"
       group: "${{ matrix.package.group }}"
       julia-version: "1.11"
     secrets: "inherit"


ChrisRackauckas-Claude:
Initial CI note: `IntegrationTest` has the expected startup failure because the current shared `SciML/.github/.github/workflows/downstream.yml@v1` does not yet declare this PR's `upstream-subdirs` input. The caller can start only after SciML/.github#115 is merged and the `v1` tag is advanced. This is why the PR body includes the exact local reusable-workflow rehearsal instead of claiming hosted Integration CI passed.

The other repository workflows are queued. Recent CI runs complete in roughly 30–60 minutes, while the monorepo matrices have taken up to about 246 minutes, so I am using a conservative five-hour checkpoint.

⎿  Setting 5-hour timer for CI completion as per CLAUDE.md instructions...
   Current time: 2026-07-15 09:53 EDT
   Will check CI results at: 2026-07-15 14:53 EDT

ChrisRackauckas-Claude:
The shared downstream workflow is now released at `.github@v1` (`f8bbdfe`). The prior IntegrationTest run ended in `startup_failure` before that release and GitHub rejected both app and token rerun requests, so commit `74d108c0` retriggers the unchanged caller against the released workflow.

⎿  Setting 3-hour timer for CI completion as per CLAUDE.md instructions...
   Current time: 2026-07-16 12:52:05 EDT
   Will check CI results at: 2026-07-16 15:52:05 EDT

ChrisRackauckas-Claude:
Post-release runtime classification is complete for the corrected ModelingToolkitBase route.

The [ModelingToolkitBase/InterfaceII job](https://github.com/SciML/BoundaryValueDiffEq.jl/actions/runs/29517383302/job/87685699711) selected `lib/ModelingToolkitBase`, reached its real test runner, and emitted a substantive summary:

```
InterfaceII | 7233 passed, 0 failed, 3 errored, 4 broken
```

The log contains one occurrence of the reusable workflow's compatibility text (the script source itself), not an executed resolver-success catch, plus a real `Test Summary:`. The job therefore is not a silent no-op. It correctly remains red for genuine integration errors, including the Lotka–Volterra ForwardDiff/Float64 path and optimization-related errors. This PR's routing change is validated; those target failures should remain visible rather than be skipped or converted to warnings.

PR #569, opened 2026-07-26
Raise LinearSolve test floor for BigFloat BVPs

ChrisRackauckas-Claude:
Please ignore this draft until reviewed by @ChrisRackauckas.

## Summary

Raise the root test compatibility floor for `LinearSolve` from `"3.12"` to
`"3.87"`.

## Why

The exact current root graph at the prior floor selected LinearSolve 3.75 and
failed the official `GROUP=Misc` BigFloat test with Sparspak's `No space for
matrix element (8, 9)` error. This is the known failure documented in issue
#509; the upstream fix is available in LinearSolve 3.87.0.

With the one-line floor change, the official dispatcher resolves LinearSolve
3.87.0 and Sparspak 0.3.15. The local control passed BigFloat 7/7, Adaptivity
19/19, Initial Guess 6/6, Scalar BVP 4/4, Default Solvers 2/2, Verbose 78/78,
Public API Package Splits 12/12, and Manifolds 11/11. The existing four
Non-Vector Input broken tests are unchanged. No tests were skipped, weakened,
or converted to warnings.

`git diff --check` and repository-wide Runic `--check .` pass. This is a test
dependency floor and does not change the runtime API.

PR #541, opened 2026-07-10
Fix MultipleShooting caches for tagged AD states

ChrisRackauckas-Claude:
Ignore this PR until it has been reviewed by @ChrisRackauckas.

## Summary

- lazily initialize and reuse MultipleShooting ODE integrator caches for each nonlinear-state element type and shooting-grid size
- keep the BVP time type unchanged when ForwardDiff introduces tagged state values
- cover the previously untested serial execution path and bump BoundaryValueDiffEqShooting to 1.17.4

## Cause

MultipleShooting hoists ODE integrator allocation out of residual evaluations. The default nonlinear polyalgorithm can evaluate that residual with a new tagged `ForwardDiff.Dual` element type when it falls back to its line-search method. Reinitializing the original `Float64` ODE cache with those states throws `Float64(::ForwardDiff.Dual)`.

The existing DifferentiationInterface Jacobian caches continue using their preallocated matching ODE caches. The new fallback is only used when another caller introduces a different state element type, and it reuses the resulting cache after the first evaluation.

This is separate from #571, which only adjusts the LinearSolve test compatibility range.

## Verification

- Exact downgrade reproducer before the fix:
  - `EnsembleThreads`: `Float64(::ForwardDiff.Dual{ForwardDiff.Tag{NonlinearFunction...}})`
  - `EnsembleSerial`: the same conversion error
- Exact downgrade after the fix, Julia 1.10.11:
  - `BOUNDARYVALUEDIFFEQ_TEST_GROUP=Core`: Basic 84/84, NLLS 56/56, Orbital 40/40
- Latest compatible dependencies, Julia 1.12.6:
  - `BOUNDARYVALUEDIFFEQ_TEST_GROUP=All`: Basic 84/84, NLLS 56/56, Orbital 40/40
  - `BOUNDARYVALUEDIFFEQ_TEST_GROUP=QA`: 19/19
- The earlier implementation's benchmark run found unchanged allocation counts and statistically neutral MultipleShooting timings; this narrower implementation will be rechecked by fresh CI.
- whole-repository Runic check: passed

PR #570, opened 2026-07-26
Cache Shooting ODE integrators by eltype

ChrisRackauckas-Claude:
Please ignore this draft until reviewed by @ChrisRackauckas.

Fixes BoundaryValueDiffEqShooting multiple-shooting caches when residual AD
changes the state/time element type. The existing cache held Float64 time
storage and attempted to convert ForwardDiff dual times to Float64.

The cache is now keyed by state element type and number of shooting intervals.
For a nonmatching element type it is remade with a matching `tspan`; result
time containers are derived from the cache rather than the primal span. This
keeps the primal path unchanged and avoids losing AD information.

Clean-main reproduction and validation on Julia 1.10 with the exact all-deps
resolver:

- unmodified `master` `f41609ed` fails Basic Shooting at
  `Float64(::ForwardDiff.Dual)` (54 pass, 1 error);
- the fixed source passes the full Shooting Core group: Basic 82/82, NLLS
  56/56, Orbital 40/40; and
- repository-wide `Runic --check .` and `git diff --check` pass.

The regression bisects to `fe630c86`.

### DIFF
--- a/lib/BoundaryValueDiffEqShooting/src/multiple_shooting.jl
+++ b/lib/BoundaryValueDiffEqShooting/src/multiple_shooting.jl
@@ -45,6 +45,7 @@ function SciMLBase.__solve(
 
     internal_ode_kwargs = (; kwargs..., odesolve_kwargs..., save_end = true)
 
+    odecache_by_eltype = Dict{Tuple{Any, Int}, Any}()
     solve_internal_odes! = @closure (
         resid_nodes,
         us,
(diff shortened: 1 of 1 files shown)


ChrisRackauckas-Claude:
Validation scratchpad: this is intentionally separate from #569. It was tested against clean `master` f41609ed with the exact all-dependency resolver; it eliminates the reproduced ForwardDiff Dual-to-Float64 error across Basic, NLLS, and Orbital Core groups.

ChrisRackauckas-Claude:
Closing as a duplicate of #541. The source patch is byte-identical; fresh clean-master f41609ed validation is recorded on #541.

PR #571, opened 2026-07-27
Allow LinearSolve 5 in root tests

ChrisRackauckas-Claude:
Please ignore this draft until reviewed by @ChrisRackauckas.

## Summary

- Allow LinearSolve 4 and 5 in the root test compatibility range.
- Unblock the BoundaryValueDiffEq downstream lane in SciML/NonlinearSolve.jl#1100.
- Keep the change test-only: LinearSolve is an extra/target here, not a runtime dependency of the root package.

## Diagnosis

The failing downstream job developed current NonlinearSolveFirstOrder, which requires LinearSolve 5, while the BoundaryValueDiffEq root test project restricted LinearSolve to major 3. A clean resolver replay reproduced the empty intersection.

A resolver bisect from NonlinearSolve commit c802a160 to current master identified 6a7687f5 (SciML/NonlinearSolve.jl#1039) as the first commit that exposes the stale ceiling. That NonlinearSolve change legitimately needs LinearSolve 4.3 or newer for its linear-solve workspace behavior, so the owner fix is to broaden this test-only BoundaryValueDiffEq range rather than revert NonlinearSolve.

## Local validation

- Exact downstream replay with current NonlinearSolve master paths (1d51c8dac), NonlinearSolveFirstOrder 2.2.4, and LinearSolve 5.2.0: GROUP=All Pkg.test() passed; observed root totals were 163 pass and 4 unchanged pre-existing broken cases, with outer Pkg.test reporting tests passed.
- Clean one-line tree with no path dependencies in Project.toml independently resolved LinearSolve 5.2.0, NonlinearSolveFirstOrder 2.2.2, NonlinearSolveBase 2.37.0, and SciMLJacobianOperators 0.1.16; GROUP=All Pkg.test() passed with the same observed root totals and outer success.
- Whole-repository Runic check passed.
- git diff --check passed.


ChrisRackauckas-Claude:
Diagnostic scratchpad:

1. Reproduced the NonlinearSolve #1100 BoundaryValueDiffEq/All failure on clean BoundaryValueDiffEq master before tests: NonlinearSolveFirstOrder 2.2.4 fixed LinearSolve to major 5 while the root test Project allowed only 3.87.
2. Resolver-bisected NonlinearSolve from c802a160 to current master. First exposing commit: 6a7687f542cde52b4bfeef5b473c6156c8d453a1 (#1039), which intentionally raises the LinearSolve floor for the quasi-Newton linear workspace.
3. Kept the owner patch to the single test-only compat line; no source or runtime dependency changes.
4. Ran both the exact current-master path replay and a second unpolluted registered-dependency GROUP=All run. Both outer Pkg.test processes exited successfully; Runic and diff checks also passed.

PR #576, opened 2026-07-30
Fix MultipleShooting caches for tagged AD states

ChrisRackauckas-Claude:
## Summary
- look up the internal ODE integrator through a `GeneralLazyBufferCache` so a matching-typed cache is built lazily for each state type MultipleShooting AD evaluations produce
- retain the primal cache for primal states; lazily built caches are sized for `maximum(all_nshoots)` so a single entry covers every grid-coarsening stage
- cover the serial MultipleShooting path alongside the existing threaded AD coverage, plus an internal testset exercising tagged-state cache matching and reuse

## Investigation
- Reproduced on unmodified `origin/master` (`219b5021`): Core completed `54 passed, 1 error` with `Float64(::ForwardDiff.Dual)` in `multiple_shooting.jl`.
- The same direct BVP reproduction also fails under a preserved July 21 lock with ForwardDiff 1.4.1, BoundaryValueDiffEqCore 2.7.3, and SciMLBase 3.36.1.
- Git history identifies `b07b168c` (`Hoist allocations out in Multiple Shooting`, 2023-11-03) as the introduction of the reusable primal cache. This is a latent package bug, not a new dependency-resolution regression.
- The bug is still live internally — tagged `us` into the primal cache errors with `MethodError` today — but current DI/NonlinearSolve internals no longer route tagged states through the loss function, so the public suite is green on master without the fix. Latent triggers remain (e.g. `AutoPolyesterForwardDiff` falls through `__cache_trait` to `NoDiffCacheNeeded`, leaving its Jacobian caches Float64-typed; currently masked by an unrelated upstream `prepare_jacobian` failure), so the guard plus the white-box `"tagged-state ODE caches"` testset keep the class of failure covered (errors without the fix, passes with it, on both `EnsembleSerial` and `EnsembleThreads`).

## Validation
- `BOUNDARYVALUEDIFFEQ_TEST_GROUP=Core` targeted runs on Julia 1.12: Basic 93/93 (incl. tagged-state ODE caches), NLLS 56/56
- QA group: 20/20 + reexport surface 52/52
- Runic format, typos, `git diff --check`

🤖 Generated with Devin CLI 3000.10.21 (model: swe-2-max) — local transcript `/home/crackauc/.local/share/devin/cli/transcripts/gold-pineapple.json`

### DIFF
--- a/lib/BoundaryValueDiffEqShooting/Project.toml
+++ b/lib/BoundaryValueDiffEqShooting/Project.toml
@@ -1,7 +1,7 @@
 name = "BoundaryValueDiffEqShooting"
 uuid = "ed55bfe0-3725-4db6-871e-a1dc9f42a757"
 authors = ["Qingyu Qu <erikqqy123@gmail.com>"]
-version = "1.17.3"
+version = "1.17.4"
 
 [deps]
 ADTypes = "47edcb42-4c32-4615-8424-f2b9edc5f35b"
--- a/lib/BoundaryValueDiffEqShooting/src/multiple_shooting.jl
+++ b/lib/BoundaryValueDiffEqShooting/src/multiple_shooting.jl
@@ -45,6 +45,7 @@ function SciMLBase.__solve(
 
     internal_ode_kwargs = (; kwargs..., odesolve_kwargs..., save_end = true)
 
+    odecache_by_eltype = Dict{Tuple{Any, Int}, Any}()
     solve_internal_odes! = @closure (
         resid_nodes,
         us,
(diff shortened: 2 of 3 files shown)


ChrisRackauckas:
Seems like this should use a lazybuffercache?

ChrisRackauckas-Claude:
Done — the `(eltype, nshoots)` dict is replaced with `PreallocationTools.GeneralLazyBufferCache`. `__multiple_shooting_matching_odecache` returns the passed-in cache when its state eltype already matches `us` (primal path unchanged), and otherwise `odecache_for_states[us]` lazily builds an integrator from `us[1:N]` sized for `maximum(all_nshoots)`, so one lazily-built entry covers every grid-coarsening stage for both `EnsembleSerial` and `EnsembleThreads`. Added an internal testset asserting the tagged cache is built on first use, reused after, and that feeding tagged states into the primal cache errors.

🤖 Generated with Devin CLI 3000.10.21 (model: swe-2-max)

Issue #560, opened 2026-07-13
Nested FIRK NLLS underconstrained cases hit SingularException on master


ChrisRackauckas-Claude:
On current `origin/master`, one nested FIRK NLLS underconstrained case throws a
singular linear solve error. This also shows up in PR CI for #559 in the NESTED
FIRK job, but the failure reproduces on clean master.

Clean master checked:

```text
4dfc3ef2872deace50764f53e8727cfbe7623d99
```

Local reproducer:

```bash
env JULIA_NUM_THREADS=1 \
    JULIA_PKG_PRECOMPILE_AUTO=no \
    JULIA_PKG_SERVER_REGISTRY_PREFERENCE=eager \
    timeout 3600 /home/crackauc/.juliaup/bin/julia +1 --startup-file=no \
    --project=lib/BoundaryValueDiffEqFIRK \
    -e 'import Pkg; Pkg.instantiate(); Pkg.build(); using SciMLBase; lines = readlines("lib/BoundaryValueDiffEqFIRK/test/nested/nlls_tests.jl"); include_string(Main, join(lines[1:179], "\n")); prob = UnderconstrainedProbArr[1]; solver = RadauIIa5(; nlsolve = NewtonRaphson(), nested_nlsolve = true); sol = solve(prob, solver; verbose = false, dt = 0.1, abstol = 1.0e-1, reltol = 1.0e-1); @show sol.retcode; @show SciMLBase.successful_retcode(sol.retcode)'
```

Observed local error:

```text
ERROR: SingularException(90)
Stacktrace:
  [9] \(A::Matrix{Float64}, B::Matrix{Float64})
      @ LinearAlgebra
 [10] nonlinearsolve_forwarddiff_solve(...)
      @ NonlinearSolveBaseForwardDiffExt .../NonlinearSolveBaseForwardDiffExt.jl:181
 [31] __internal_solve
      @ BoundaryValueDiffEqCore/src/default_internal_solve.jl:107
 [32] __perform_firk_iteration(...)
      @ BoundaryValueDiffEqFIRK/src/firk.jl:579
```

Recent PR CI signature from #559 NESTED job:

```text
Solver: LobattoIIIa4 with NewtonRaphson: Error During Test
  at lib/BoundaryValueDiffEqFIRK/test/nested/nlls_tests.jl:201
  LinearAlgebra.SingularException(72)
...
The operation was canceled.
```

[...]

PR #579, opened 2026-07-30
Remove FIRK dependency blanket reexports

ChrisRackauckas-Claude:
Ignore this PR until it has been reviewed by @ChrisRackauckas.

## Summary

- remove FIRK's blanket reexports of ADTypes, BoundaryValueDiffEqCore, and SciMLBase
- retain a fixed, reviewed facade of only `BVProblem`, `TwoPointBVProblem`, and `solve`
- import dependency APIs from their owning modules in FIRK source and tests
- complete definition-site SciMLStyle docstrings for all 17 FIRK algorithms and render them in the manual
- keep caches and implementation helpers private
- remove Reexport and add the direct NonlinearSolveFirstOrder dependency without changing FIRK's 1.17.3 version

## Reexport proof

The rendered FIRK solver page uses only:

```julia
using BoundaryValueDiffEqFIRK
```

Its executed doctest constructs both a `BVProblem` and a `TwoPointBVProblem`, then calls `solve` for each. Those are therefore the complete retained dependency-owned workflow facade. `FIRK_REEXPORTS` and strict QA use the same fixed three-name allowlist; no blanket or generated list remains.

The audited allowlist changes from 274 dependency reexports to 3: the three names above are retained and the other 271 former ADTypes/BoundaryValueDiffEqCore/SciMLBase exports are removed.

## API and metadata audit

- all 17 FIRK algorithm exports retain definition-site `"""..."""` plus `Base.@kwdef struct` docstrings
- all 17 are rendered through the FIRK manual's `@docs` blocks
- every pre-existing `[sources]` entry is preserved exactly
- SciMLTesting compat is `2.4`; final isolated QA resolved 2.7.0 with standard strict defaults
- FIRK remains version 1.17.3; no umbrella/docs/test compat admits FIRK 2

## Local verification

- root full suite, release: 163 passed, 4 pre-existing broken
- root full suite, Julia 1.10: 163 passed, 4 pre-existing broken
- FIRK Core, release: 226/226 passed
- FIRK Core, Julia 1.10: 226/226 passed
- FIRK non-AD/non-underconstrained group matrix, release: 617 passed, 46 pre-existing broken
- FIRK non-AD/non-underconstrained group matrix, Julia 1.10: 617 passed, 46 pre-existing broken
- FIRK AD, Julia 1.10: 9/9 passed
- strict FIRK QA, release: 20/20 passed
- strict FIRK QA, Julia 1.10: 18/18 passed
- rendered docs and doctests: passed on release and Julia 1.10
- focused FIRK facade regression, release via `Pkg.test()`: 5/5 passed
- focused FIRK facade regression, Julia 1.10: 5/5 passed
- export smoke assertions: passed on release and Julia 1.10
- all 121 tracked Julia files: Runic 1.7.0 passed
- `git diff --check`: passed

The release AD group and Julia 1.13 prerelease AD group abort in unchanged Enzyme-backed code on clean master as well; the prerelease reproduction used Enzyme 0.13.199 and faulted at `firk.jl:99`. An adjacent Enzyme 0.13.198 full-sequence run avoided that segfault but exposed separate Mooncake/Julia 1.13 errors. The separate repair/history is #547. The nested underconstrained group has the existing clean-master `SingularException` tracked in #560 and upstream in SciML/NonlinearSolve.jl#1054. Neither repair is included here.

The clean-master downgrade floor repairs #596 and #597 and the one-line Core patch-version repair #599 have merged independently into current master. None of those repair changes is part of this PR commit.


ChrisRackauckas:
## Rework: the reexport removal was a silent breaking change

As written, this PR deleted `@reexport using ADTypes, BoundaryValueDiffEqCore, SciMLBase`,
which removed **274 names** from `using BoundaryValueDiffEqFIRK`. Nothing inside the repo
depends on that facade, so CI stayed green while downstream `using BoundaryValueDiffEqFIRK`
code lost `BVProblem`, `solve`, `AutoForwardDiff`, `BVPJacobianAlgorithm`, `NewtonRaphson`,
and the rest.

Reworked to the shape of SciML/Sundials.jl#553: keep the blanket reexport out, but bring
the same bindings back as an explicit `using X: ...` / `export ...` pair, and restore the
`reexports_allow` / `api_docs_kwargs` whitelisting in `lib/BoundaryValueDiffEqFIRK/test/qa/qa.jl`.

### How the list was derived

Generated, not hand-written, in the sublibrary's own environment:

```julia
filter(s -> Base.isexported(M, s), names(M))   # M = ADTypes, BoundaryValueDiffEqCore, SciMLBase
```

`names(M)` alone is wrong on Julia >= 1.11 (it also reports `public`-but-not-exported names,
603 across the three modules); `Reexport` only re-exports the exported ones. Imports are
grouped by the module each name is exported from — ADTypes (38), BoundaryValueDiffEqCore
minus what it reexports from SciMLBase (51), SciMLBase (185) — which is why
`check_all_explicit_imports_via_owners` and `check_all_explicit_imports_are_public` stay
green without new ignores (both discard names that are public in the module they are
imported from), and `check_no_stale_explicit_imports` does too (a re-exported import is
never stale).

The 274 names match `FIRK_REEXPORTS` in `test/qa/reexports.jl` exactly — empty set
difference in both directions.

### Verification

- `names(BoundaryValueDiffEqFIRK)`: **292 on `master`, 292 on this branch**, `setdiff` empty
  in both directions, and every exported symbol `isdefined` in the module.
- QA lane: 20/20 passing.
- Smoke solve (`RadauIIa5` on a two-point BVP) with `BoundaryValueDiffEqFIRK`, `SciMLBase`,
  `ADTypes`, and `BoundaryValueDiffEqCore` all loaded together: `ReturnCode.Success`, no
  ambiguity warnings (the re-exports are the same bindings, so double-`using` is fine).
- Runic: no change.

Diffstat moved from +21/-13 to +151/-5, the expected signature — the stripped names come
back as explicit exports rather than disappearing.

The `Reexport` dependency removal and the added explicit `using` lines in the test files
are kept as-is; they are correct independent of the facade.

The root `BoundaryValueDiffEq` package is unaffected: it already imports from FIRK
explicitly (`using BoundaryValueDiffEqFIRK: BoundaryValueDiffEqFIRK, LobattoIIIa2, ...`),
so `ROOT_REEXPORTS` does not change.

oscardssmith:
probably worth a breaking change for sub-libraries still using reexport (no break needed at the OrdinaryDiffEq layer)

PR #580, opened 2026-07-30
Remove BoundaryValueDiffEqShooting blanket reexports

ChrisRackauckas-Claude:
## Summary
- remove the explicit restoration of the old ADTypes, BoundaryValueDiffEqCore, and SciMLBase forwarding surface; BoundaryValueDiffEqShooting exports only `Shooting` and `MultipleShooting`
- remove the reexport allowlist and public-doc/rendered-doc QA exemptions
- enable strict root doctests and checkdocs: no `warnonly`, direct owner links for SciMLBase APIs, and executable Shooting documentation examples
- retain the direct owner imports introduced in #577/#579; no solver implementation or Core tests were changed

Undocumented dependency reexports are removed as API cleanup, not a breaking documented-interface change.

## Validation
- QA: `19/19` passed, with no reexport or public-doc exceptions
- strict `julia +1.12 --project=docs docs/make.jl` exited 0 with doctests, cross-references, CheckDocument, and HTML rendering enabled; no `warnonly` setting remains
- Runic check and `git diff --check` passed
- `BOUNDARYVALUEDIFFEQ_TEST_GROUP=Core ... Pkg.test()` fails identically here and on untouched master `219b5021`: `54 passed, 1 errored` from the existing `Float64(::ForwardDiff.Dual)` MultipleShooting failure. No test was changed, skipped, or suppressed.


ChrisRackauckas:
### Rework: the reexport removal was a silent breaking change

`@reexport using ADTypes, BoundaryValueDiffEqCore, SciMLBase` was contributing **274
names** to `using BoundaryValueDiffEqShooting`. Deleting it dropped all of them from the
package's public surface. No code inside this repo depends on those names, so CI stayed
green — but downstream code that got `BVProblem`, `NewtonRaphson`, `TrustRegion`,
`AutoFiniteDiff`, `ReturnCode`, `solve`, ... from `using BoundaryValueDiffEqShooting`
would break. (The test-file churn in this PR — adding `using NonlinearSolveFirstOrder:
NewtonRaphson, TrustRegion` etc. — is exactly that break showing up in-repo.)

Commit d29391e follows the pattern of SciML/Sundials.jl#553: keep Reexport gone, but
bring the names back as an **explicit** import + export list of the same bindings.

- `src/BoundaryValueDiffEqShooting.jl`: `using ADTypes: ...` / `using SciMLBase: ...` /
  `using BoundaryValueDiffEqCore: ...` for the reexported API, plus a second `export`
  block listing all 274 names.
- `test/qa/qa.jl`: restores `reexports_allow = SHOOTING_REEXPORTS` and the matching
  `api_docs_kwargs` ignores. `SHOOTING_REEXPORTS` in `test/qa/reexports.jl` is the
  already-audited list and matches `filter(x -> Base.isexported(m, x), names(m; all =
  true, imported = true))` over the three modules exactly — it was not hand-written here.

Verification:

```
names(BoundaryValueDiffEqShooting)  base: 277   after: 277
setdiff(base, after) == []          setdiff(after, base) == []
```

`test/qa/qa.jl` runs green locally (19/19 passing: Aqua, all six ExplicitImports checks,
public API docs, and the no-unapproved-reexports audit). Both changed files are
Runic-clean. The `Core` group has one pre-existing `MultipleShooting` /
`ForwardDiff.Dual` error that reproduces identically on 7a1da23 (54 pass / 1 error either
way), so it is not from this commit.

Issue #556, opened 2026-07-11
Master CI: MIRK/Shooting ForwardDiff Dual cache failures and sublibrary downgrade blockers


ChrisRackauckas-Claude:
Current `master` at `906ff8948661c8a917111db3e3e5a4e7b1185ae0` still has several failures after the root QA path bug fixed in #555.

## Affected current master jobs

- Documentation run 29139451376, job 86509810722: <https://github.com/SciML/BoundaryValueDiffEq.jl/actions/runs/29139451376/job/86509810722>
  - Fails in `docs/src/tutorials/extremum.md:14-30`, `sol = solve(prob, MIRK4(), dt=0.05)`.
  - Error class: `MethodError: no method matching Float64(::ForwardDiff.Dual{...})`.
  - Relevant frame: `BoundaryValueDiffEqMIRK/src/mirk.jl:338` in `__perform_mirk_iteration`, called from `mirk.jl:293`.
- Downgrade Sublibraries run 29139451449, Shooting job 86509838701: <https://github.com/SciML/BoundaryValueDiffEq.jl/actions/runs/29139451449/job/86509838701>
  - Fails in `lib/BoundaryValueDiffEqShooting/test/Core/basic_problems_tests.jl:4`, with the failing test call at line 66.
  - Error class: `MethodError: no method matching Float64(::ForwardDiff.Dual{...})`.
  - Relevant frames include `SciMLBase.reinit!`, `BoundaryValueDiffEqShooting/src/multiple_shooting.jl:403/406`, and `__multiple_shooting_mpoint_loss!` around `multiple_shooting.jl:507`.
- Downgrade Sublibraries MIRK/FIRK jobs 86509838685 and 86509838696 currently fail earlier in resolution. CI reports `SciMLBase` is restricted by BoundaryValueDiffEqCore to `3.30.0-3.33.1` while an explicit downgrade requirement selects `SciMLBase 3.0.0`. With the current registry locally, the same class of downgrade resolver blocker appears as `LinearSolve@4.2.1` incompatible with MIRK/FIRK `LinearSolve = "2.36.2, 3"`. Raising the local MIRK/FIRK `SciMLBase`/`LinearSolve` compat floors allowed MIRK to reach tests, but then MIRK failed with the same `Float64(::ForwardDiff.Dual)` class in `Core/mirk_basic_tests.jl:311` and `:434`, so this is not just a compat-only fix.

## Local reproductions

[...]

Issue #548, opened 2026-07-10
FIRK EXPANDED master job runs past one hour locally on fcc08ec5


ChrisRackauckas-Claude:
## Summary

`master` at `fcc08ec5ed78c09662df5d90975fbacbf481b89c` has a separate FIRK EXPANDED CI failure:

- Workflow run: https://github.com/SciML/BoundaryValueDiffEq.jl/actions/runs/29097963783
- Job: `sublibraries / lib/BoundaryValueDiffEqFIRK [EXPANDED] / Julia 1 / Tests - EXPANDED (Julia 1)`
- Job id: `86379299251`
- GitHub reports the job completed failure at `2026-07-10T14:40:49Z`.

At the time of investigation, the overall workflow run still reported `status: in_progress`, so both `gh run view --job 86379299251 --log` and `/actions/jobs/86379299251/logs` did not provide logs (`gh` says logs will be available when the run completes; direct job logs returned HTTP 404).

## Local reproduction attempt

Clean checkout path:

`/home/crackauc/sandbox/tmp_20260708_121627_10236/firk_ad_master_investigation_clean`

Command:

```sh
timeout 3600 env BOUNDARYVALUEDIFFEQ_TEST_GROUP=EXPANDED \
  /home/crackauc/.juliaup/bin/julia +1.12 \
  --project=lib/BoundaryValueDiffEqFIRK \
  -e 'using Pkg; Pkg.test()' \
  > ../firk_expanded_repro_master_fcc08ec5.log 2>&1
```

Result: exited `124` from the one-hour timeout. The test process was CPU-bound throughout. The log showed `Testing Running tests...` and many `@time` solve lines from the expanded basic tests, but no Julia exception, assertion failure, or segfault before timeout.

## Classification

This does not look like the FIRK AD segfault from the AD job. The AD job reproduced as `signal 11` through Enzyme forward mode at `lib/BoundaryValueDiffEqFIRK/src/firk.jl:1525` / `__firk_mpoint_jacobian!`; that is addressed separately in draft PR #547.

The EXPANDED symptom currently looks like a long-runtime/nontermination issue in the EXPANDED suite. It still needs run logs once GitHub releases them, then a narrower reproduction/bisect of the specific expanded test case.

ChrisRackauckas-Claude:
## Still reproduces on current master — now ~2h21m, and it is not a hang

[...]

Issue #535, opened 2026-07-08
MIRK4 Swirling Flow III fails on default nonlinear solver with ForwardDiff Dual conversion


ChrisRackauckas-Claude:
## Reproduction

Clean checkout of `SciML/BoundaryValueDiffEq.jl` at current `origin/master`:

```bash
git clone https://github.com/SciML/BoundaryValueDiffEq.jl.git BoundaryValueDiffEq-master-failure
cd BoundaryValueDiffEq-master-failure
git fetch origin
git reset --hard origin/master
```

Observed HEAD:

```text
36e9daa5f64a7e2a45ee0256559e739f027d6f8d
36e9daa5 Merge pull request #534 from ChrisRackauckas-Claude/fix-mirk-constrained-interpolation-533
```

Run the minimal version of `lib/BoundaryValueDiffEqMIRK/test/Core/mirk_basic_tests.jl` testset `Swirling Flow III`:

```bash
timeout 3600 /home/crackauc/.juliaup/bin/julia +1.10 --project=lib/BoundaryValueDiffEqMIRK -e 'using Pkg; Pkg.instantiate(); using BoundaryValueDiffEqMIRK; eps=0.01; function swirling_flow!(du,u,p,t); eps=p; du[1]=u[2]; du[2]=(u[1]*u[4]-u[3]*u[2])/eps; du[3]=u[4]; du[4]=u[5]; du[5]=u[6]; du[6]=(-u[3]*u[6]-u[1]*u[2])/eps; return; end; function swirling_flow_bc!(res,u,p,t); res[1]=u(0.0)[1]+1.0; res[2]=u(0.0)[3]; res[3]=u(0.0)[4]; res[4]=u(1.0)[1]-1.0; res[5]=u(1.0)[3]; res[6]=u(1.0)[4]; return; end; prob=BVProblem(swirling_flow!, swirling_flow_bc!, zeros(6), (0.0,1.0), eps); sol=solve(prob, MIRK4(); dt=0.01, abstol=1.0e-4); @show sol.retcode'
```

## Result

This fails on clean current master with:

```text
ERROR: MethodError: no method matching Float64(::ForwardDiff.Dual...)
```

[...]

Issue #180, opened 2024-04-18
Precompilation triggered at each include


bmxam:
Hi,

I am trying to use the `TwoPointBVPProblem` from `BoundaryValueDiffEq`. I put my code in a file and runs it by including in the REPL. The code runs great but the first call to `solve` always triggers a (long) "recompilation" time. The next `solve` calls inside the files are ok. But if I keep the same REPL and include the file again, I always have this first precompilation time. Is there a way to avoid / improve it (apart from solution like `PackageCompiler`)?

Here is the full code and the timings on my machine. I am using `BoundaryValueDiffEq - 5.7.1`.

```julia
using BoundaryValueDiffEq

function f!(du, u, p, t)
    F0 = u[1]
    F1 = u[2]
    F2 = u[3]

    du .= F1, F2, 2 * F1 * (F2 / (F0 - p.myeps) - 1 / (p.r^2))
end

bc2a!(resid_a, u_a, p) = resid_a[1] = u_a[1]
bc2b!(resid_b, u_b, p) = resid_b .= u_b[1] + 1.0, u_b[2]

tspan = (-1.0, 0.0)
u0 = [-1.0, 0.0, 0.0] # I don't know how to use the init function

myeps = 1e-10
dt = 0.01
p = (r=1.0, myeps=myeps)
bvp = TwoPointBVProblem(f!, (bc2a!, bc2b!), u0, tspan, p; bcresid_prototype=(zeros(1), zeros(2)))
@time sol = solve(bvp, MIRK4(), dt=dt)
@time sol = solve(bvp, MIRK4(), dt=dt)
```

```julia
9.803415 seconds (2.23 M allocations: 153.588 MiB, 0.22% gc time, 99.98% compilation time: 100% of which was recompilation)
0.002087 seconds (13.48 k allocations: 1.517 MiB)
```

ChrisRackauckas-Claude:
Some measurements on this, taken while investigating #548 (master `1352f171`, Julia 1.12.6).

The latency you are seeing is one full specialization of the BVP solve stack per `(problem type, solver type)` pair. Solving a small linear BVP, with a freshly defined closure type for each measurement:

| Measurement | Time |
| --- | --- |
| First solve, default options | 36.1 s |
| Same problem and solver solved again in-process | 0.0 s |
| New problem type, same solver | 19.2 s |
| Known problem type, different solver stage | 18.1 s |

[...]

PR #583, opened 2026-08-02
Split FIRK EXPANDED tests into five CI groups (#548)

ChrisRackauckas-Claude:

Fixes the CI symptom in #548 by splitting the FIRK `EXPANDED` test group. Also includes the measurements explaining *why* it is slow, which point at a library-level latency problem worth fixing separately.

## The problem

`EXPANDED` takes **2h21m45s** as a single job locally (master `1352f171`, Julia 1.12.6). It does not hang — the one-hour CI/local timeout was cutting it off mid-run, which is why #548 had no usable log.

Measured breakdown of the old single group:

| Testset | Time |
| --- | --- |
| FIRK Expanded Basic Tests | 79m40s |
| FIRK Expanded NLLS Tests | 40m31s |
| FIRK Expanded Singular BVP Tests | 11m12s |
| FIRK Expanded Ensemble Tests | 9m09s |
| FIRK Expanded VectorOfVector Initials Tests | 1m08s |

## Why it is slow: it is all compilation

Timing solves of a small linear BVP, each with a freshly defined closure type:

| Measurement | Time |
| --- | --- |
| First solve, default options | 36.1 s |
| **Same problem and solver solved again in-process** | **0.0 s** |
| New problem type, same solver | 19.2 s |
| Known problem type, different stage (`RadauIIa3`) | 18.1 s |

The numerical work is free. The cost is one full specialization of the FIRK → NonlinearSolve → AD stack per `(problem type, solver type)` pair, at ~20s each. `Affineness` and `Convergence on Linear` each sweep 16 solvers over 4 problems, so they pay it 64 times apiece — ~50 minutes for those two testsets alone.

This is not FIRK-specific. MIRK is the same (`MIRK4` 24.2s, `MIRK6` 21.2s per fresh problem type); FIRK EXPANDED just has the largest solver × problem cross product. It is also the same root cause as #180.

## Where the compile time goes

The default `nlsolve` is `__FastShortcutBVPCompatibleNonlinearPolyalg`, a 3-algorithm polyalgorithm. `__generated_polysolve` unrolls it, so all three get compiled even though `NewtonRaphson` (the first) normally converges and the other two never run:

| `nlsolve` | Compile time |
| --- | --- |
| bare `NewtonRaphson` | 12.5 s |
| PolyAlgorithm, 1 alg | 11.3 s |
| PolyAlgorithm, 2 algs | 18.2 s |
| PolyAlgorithm, 3 algs (current default) | 23.3 s |

The wrapper itself is free; each extra fallback costs ~5–7 s **per specialization**. Roughly half of the suite's runtime is compiling nonlinear-solver fallbacks that are never executed. Sparse-vs-dense AD (12.7 → 13.0 s) and `adaptive` (13.0 → 11.8 s) are noise by comparison.

**This is not addressed here** — deferring fallback compilation until the primary solver actually fails is a change to solver defaults and belongs in its own PR. Filing the numbers so the decision can be made deliberately.

## What this PR does

Splits `EXPANDED` into five matrix groups, extracting the shared problem set into `expanded/firk_test_setup.jl` and giving the two expensive solver sweeps their own files:

| New group | Contents | Time | Result |
| --- | --- | --- | --- |
| `EXPANDED_BASIC` | pendulum, interpolation, swirling flow, continuation, parameter estimation, initial guess | 30m01s | 126 pass |
| `EXPANDED_AFFINENESS` | `Affineness` sweep | 25m06s | 64 pass |
| `EXPANDED_CONVERGENCE` | `Convergence on Linear` sweep | 25m44s | 44 pass, 20 broken |
| `EXPANDED_NLLS` | `nlls_tests.jl` | 39m60s | 24 pass |
| `EXPANDED_MISC` | ensemble, singular BVP, vectorofvector | ~20m | 56 pass, 1 error (pre-existing, see below) |

Longest job: **2h21m → 30m**.

Test counts are conserved exactly — the old basic-tests file ran 254 (234 pass + 20 broken); the split runs 126 + 64 + 64 = 254 with the same pass/broken split.

`runtests.jl` keeps an aggregate `EXPANDED` group so the whole formulation can still be run in one process locally. Like `Core`, it is deliberately not a matrix entry in `test_groups.toml` and not in `all`, so nothing runs twice.

## Verification

Each of the five groups was run locally to completion via `Pkg.test()` on Julia 1.12.6 with `BOUNDARYVALUEDIFFEQ_TEST_GROUP` set:

```
EXPANDED_BASIC        exit 0    FIRK Expanded Basic Tests        126  126  30m01.0s
EXPANDED_AFFINENESS   exit 0    FIRK Expanded Affineness Tests    64   64  25m05.8s
EXPANDED_CONVERGENCE  exit 0    FIRK Expanded Convergence Tests   44   20   64  25m44.4s
EXPANDED_NLLS         exit 0    FIRK Expanded NLLS Tests          24   24  39m59.5s
EXPANDED_MISC         exit 1    ensemble 28/28, singular 28/28, vectorofvector 1 error
```

Runic is clean over `lib/BoundaryValueDiffEqFIRK/`.

`EXPANDED_MISC` exits 1 on a **pre-existing** failure that this PR does not touch and does not fix: `expanded/vectorofvector_initials_tests.jl:71` errors with `MethodError: no method matching Float64(::ForwardDiff.Dual)` from `firk.jl:1403`, where `eval_sol.u[1:end] .= y_` writes Duals into a plain `Vector{Vector{Float64}}`. It reproduces identically on unmodified master (details in https://github.com/SciML/BoundaryValueDiffEq.jl/issues/548#issuecomment-5154191889); the nested formulation's equivalent test passes. Same family as #535 and #556. Happy to split it into its own issue.

## Not done here

- `NESTED` was not split. Partial timings (`ensemble` 11m28s, `vectorofvector` 1m42s) are well inside budget; the `nested/firk_basic_tests.jl` and `nlls_tests.jl` runs had not finished when this was opened, so splitting it should be a follow-up if the numbers warrant.
- The polyalgorithm latency fix described above.

https://claude.ai/code/session_018aqZLpyNrFJBVfpecaEStp


ChrisRackauckas-Claude:
CI results: 13 of 15 EXPANDED jobs pass, all completing in 18-26 min instead of timing out.

The two failure classes:

**`EXPANDED_MISC` (all three versions)** - the pre-existing `vectorofvector_initials_tests.jl:71` Dual bug described in the PR body. Reproduces on unmodified master; not introduced or fixed here.

**`EXPANDED_BASIC` on lts** - not a test failure. Its `Run tests` step has conclusion `null` rather than `failure`, no steps after it ran, and the job log returns `BlobNotFound`. On the stacked #584 run, which contains this same commit, `EXPANDED_BASIC` passes on lts (21m51s) and fails on Julia 1 instead, with the identical `null`-step / no-log signature. The affected version moves between runs, so this is the runner being killed mid-step rather than anything in the tests.

Worth investigating separately: locally this group's `Pkg.test` process peaked at 15.4 GB RSS against a 16 GB standard runner, and its tests allocate 244 GiB over 3.85 G allocations. I could not confirm OOM directly because the logs that would show it are the ones that were not retained.

I could not reproduce on lts locally: Julia 1.10 fails to resolve this package set on my machine (`BracketingNonlinearSolve depends on ChainRulesCore, but no such entry exists in the manifest`) before any test runs, with a clean manifest, a fresh depot, and an isolated checkout alike.

PR #584, opened 2026-08-02
Split FIRK NESTED tests into seven CI groups

ChrisRackauckas-Claude:

Stacked on #583 (contains its commit; merge that first). Same treatment for the `NESTED` group, which turned out to be in worse shape than `EXPANDED`.

## What was wrong

Timed per file locally on master (Julia 1.12.6):

| File | Time |
| --- | --- |
| `nested/firk_basic_tests.jl` | 1h50m |
| `nested/nlls_tests.jl` | **hit the 3h timeout without finishing** |
| `nested/ensemble_tests.jl` | 11m28s |
| `nested/vectorofvector_initials_tests.jl` | 1m42s |

Two distinct causes, neither the same as #548's:

**1. `Simple Pendulum` is 55m of real solve time.** It runs 17 solvers at `dt = 0.005` over `(0, π/2)`, and nested solvers do an inner nonlinear solve per mesh point. The expanded equivalent is 8m51s. This is not compilation.

**2. `Underconstrained BVP` in the NLLS file is pathologically slow.** The 3h timeout was *not* a hang — a kill-time backtrace showed it inside `FIRK_nlsolve!` (`collocation.jl:149`), actively solving, with 164 GB allocated and 89850 GCs. Timing the cases individually:

```
prob=1  RadauIIa5 with NewtonRaphson      250.9 s   retcode=Success
prob=1  LobattoIIIa4 with NewtonRaphson  1230.6 s   retcode=Success
prob=1  LobattoIIIb4 with NewtonRaphson  > 29 min   (still running)
```

They succeed; they just take 4–20+ minutes each, and there are 22 of them. **This PR does not fix that** — it is a real solver problem, plausibly related to #560, and wants its own investigation. What this PR does is stop it from taking the rest of the NLLS coverage down with it.

## What this PR does

Splits `NESTED` into seven groups, mirroring #583's structure plus a group for the pendulum sweep and one for underconstrained NLLS:

| New group | Time | Result |
| --- | --- | --- |
| `NESTED_BASIC` | 20m28s | 100 pass |
| `NESTED_AFFINENESS` | 25m36s | 68 pass |
| `NESTED_CONVERGENCE` | 24m21s | 38 pass, 26 broken |
| `NESTED_PENDULUM` | 57m27s | 17 pass |
| `NESTED_NLLS` | 24m41s | 48 pass |
| `NESTED_MISC` | ~14m | 30 pass |
| `NESTED_NLLS_UNDERCONSTRAINED` | hours | not run to completion |

The headline: **`NESTED_NLLS` goes from timing out at 3h to passing in 25 minutes**, because the overconstrained coverage no longer waits behind the underconstrained sweep.

`NESTED_PENDULUM` stays long at 57m because the cost is genuine solve time; its `timeout` is set to 180 accordingly. `NESTED_NLLS_UNDERCONSTRAINED` gets 360 and is expected to remain over budget until the underlying slowness is addressed — that is deliberate, so the problem stays visible instead of being hidden inside a group that used to fail for unclear reasons.

Shared setup is extracted into `nested/firk_test_setup.jl` and `nested/nlls_test_setup.jl`. As in #583, `runtests.jl` keeps an aggregate `NESTED` group for local use that is not a matrix entry.

## Verification

Each of the six terminating groups was run to completion locally via `Pkg.test()` with `BOUNDARYVALUEDIFFEQ_TEST_GROUP` set, all exit 0:

```
NESTED_BASIC        exit 0   FIRK Nested Basic Tests            100  100  20m28.4s
NESTED_AFFINENESS   exit 0   FIRK Nested Affineness Tests        68   68  25m36.2s
NESTED_CONVERGENCE  exit 0   FIRK Nested Convergence Tests       38   26   64  24m20.6s
NESTED_PENDULUM     exit 0   FIRK Nested Simple Pendulum Tests   17   17  57m27.2s
NESTED_NLLS         exit 0   FIRK Nested NLLS Tests              48   48  24m40.7s
NESTED_MISC         exit 0   ensemble 28/28 12m55.2s, vectorofvector 2/2 1m34.5s
```

Test counts are conserved: the old basic file's 68 + 64 + 17 + the remaining testsets equals 68 + 64 + 17 + 100 across the new groups, with the same pass/broken split. `NESTED_NLLS_UNDERCONSTRAINED` was deliberately not run to completion; its content is unchanged from what `nlls_tests.jl` ran before.

Runic is clean over `lib/BoundaryValueDiffEqFIRK/`.

## Follow-ups this does not do

- The underconstrained NLLS slowness above.
- The compile-latency work quantified in #583 and https://github.com/SciML/BoundaryValueDiffEq.jl/issues/180#issuecomment-5154471409 (default 3-algorithm polyalgorithm compiling two never-executed fallbacks, ~50% of per-solve compile time).

https://claude.ai/code/session_018aqZLpyNrFJBVfpecaEStp


ChrisRackauckas-Claude:
CI results are in and the NESTED split works: **every NESTED_\* group passes on all three Julia versions**, including `NESTED_NLLS`, which previously hit the 3h timeout.

| Group | Julia 1 | lts | pre |
| --- | --- | --- | --- |
| `NESTED_BASIC` | 15m18s | 16m44s | 19m55s |
| `NESTED_AFFINENESS` | 22m28s | 13m42s | 19m45s |
| `NESTED_CONVERGENCE` | 18m44s | 15m33s | 17m33s |
| `NESTED_MISC` | 13m34s | 11m28s | 16m45s |
| `NESTED_NLLS` | 19m54s | 19m39s | 22m25s |
| `NESTED_PENDULUM` | 38m21s | 45m2s | 34m23s |

`NESTED_NLLS_UNDERCONSTRAINED` is still queued; it is expected to run long, which is the point of isolating it.

### On the `EXPANDED_BASIC` failures

`EXPANDED_BASIC` failed on lts in #583's run and on Julia 1 in this run, passing on the other two versions both times. It is not a test failure:

- The `Run tests` step has conclusion `null` rather than `failure` in both cases, and no steps after it ran.
- The job logs return `BlobNotFound` for both.
- The version affected changes between runs.

That signature is the runner being killed mid-step rather than a test failing. Memory is the plausible cause and worth a look: locally the `Pkg.test` process for this group peaked at 15.4 GB RSS, against 16 GB on a standard runner, and the basic tests allocate 244 GiB across 3.85 G allocations. I have not confirmed OOM directly, since the logs that would say so are exactly the ones GitHub did not retain.

PR #521, opened 2026-07-02
Add mass matrix support for FIRK and MIRK

SebastianM-C:
via unprojected collocation for index-1 DAEs based on Ascher & Spiteri (1994)

## Checklist

- [x] Appropriate tests were added
- [x] Any code changes were done in a way that does not break public API
- [x] All documentation related to code changes were updated
- [x] The new code follows the
  [contributor guidelines](https://github.com/SciML/.github/blob/master/CONTRIBUTING.md), in particular the [SciML Style Guide](https://github.com/SciML/SciMLStyle) and
  [COLPRAC](https://github.com/SciML/COLPRAC).
- [x] Any new documentation only uses public API

PR #589, opened 2026-08-09
Make the optimize-path constraint Jacobian structural and self-consistent

SebastianM-C:
**Stacked on #521.** This branch is `smc/mass_matrix` plus two commits, so the diff here
currently includes all of #521. Only the top two commits are new:

- `Make the optimize-path constraint Jacobian structural and self-consistent`
- `Evaluate algebraic constraints at the full stacked endpoint in the OC path`

Once #521 merges I'll rebase onto master and the diff will shrink to just those. Review
#521 first.

## The problem

On the optimization (BVP-with-cost) path, `OptimizationFunction` was handed the
nlsolve-oriented `cons_j` and `cons_jac_prototype = sparse(jac_prototype)` pair. That
prototype is materialized *numerically at the initial guess*, so entries that happen to
vanish there — e.g. the control columns of an optimal-control problem — are dropped from
the pattern. The refill then follows the DI-detected pattern and tries to write the
missing entries, breaking the frozen-nonzero contract the interior-point backends rely on.

The nlsolve path re-detects sparsity whenever the mesh changes, so it never hit this. The
optimization backends freeze the pattern once and refill it every iterate, so the pattern
has to be **structural** (input-independent) rather than local.

## Changes

- Hand `OptimizationBase` only the assembled `loss`, so it derives the constraint-Jacobian
  prototype and the Jacobian values from a single DI preparation and their nonzero
  patterns agree by construction.
- Add `__optimization_sparsity_detector`, which promotes a `TracerLocalSparsityDetector`
  to the global `TracerSparsityDetector` on this path.
- Evaluate algebraic constraints at the full stacked endpoint in the OC path.

## A latent bug this surfaced

Making the pattern structural means the constraint residual is traced with
`SparseConnectivityTracer` duals, which `mul!` cannot handle (it branches on `iszero`,
undefined for global tracers). So `__maybe_matmul!` needs a non-`mul!` fallback for
non-float element types.

That fallback already existed, but was **unreachable**: its `z` was untyped and therefore
less specific than the `z::AbstractArray` fast path, so array arguments always went to
`mul!`. Narrowing the fast path to float element types made the loop live for the first
time — and it was broken. It wrote

```julia
@. z = α * A[:, j] * b[j]
```

where the `@views`-rewritten column slice is *itself broadcast* inside `@.`, giving the
wrong shape. Binding the slice before the broadcast fixes it. It showed up as
`DimensionMismatch: cannot broadcast array to have fewer non-singleton dimensions` in the
multi-point interpolation tests, whose boundary conditions call `sol(t)` during sparsity
detection.

## Testing

Measured against `smc/mass_matrix` as the baseline, so these isolate the two new commits:

| Testset | base | this PR |
|---|---|---|
| Interpolation | 80 pass | 80 pass |
| Rocket launching problem | 2 pass | 2 pass |
| Constrained interpolation without control variables | 4 pass | 4 pass |
| Simple index-1 DAE | 30 pass | 30 pass |
| Ascher & Spiteri example problem 1 | 8 pass | 8 pass |
| Mesh adaptivity is not supported for DAEs | 2 pass | 2 pass |

## Checklist

- [x] Appropriate tests were added
- [x] Any code changes were done in a way that does not break public API
- [ ] All documentation related to code changes were updated
  [COLPRAC](https://github.com/SciML/COLPRAC).
- [x] Any new documentation only uses public API


lib/BoundaryValueDiffEqCore/src/internal_problems.jl
```diff
@@ -98,11 +98,29 @@ function integral(fun, domain)
...
+
+ADtype handed to `OptimizationFunction` on the BVP-with-cost path. The explicit
+`SecondOrder` core keeps optimizers that need Hessians (e.g. Ipopt) from warning and
+wrapping the first-order ADtype themselves.
+
+Only the assembled constraint function `cons` is passed alongside it, never the
+nlsolve-oriented `cons_j`/`cons_jac_prototype` pair: OptimizationBase then derives the
+constraint-Jacobian prototype and its values from a single DI preparation, so their
+nonzero patterns agree by construction. The nlsolve `jac_prototype` is materialized
+numerically at the initial guess, so entries that vanish there (e.g. the control columns
```
ChrisRackauckas: Why would we do this instead of using the optimized path?

